PR #90, opened 2019-05-16
add codecov to Circle CI

bobby-brennan:


.circleci/config.yml
```diff
@@ -50,6 +57,7 @@ jobs:
       - run: go get -u golang.org/x/lint/golint
       - run: go list ./... | grep -v vendor | xargs golint -set_exit_status
       - run: go list ./... | grep -v vendor | xargs go vet
+      - *update_coverage
```
robscott: This seems like something we'd only want to run on commits to the master branch.

bobby-brennan: Codecov appears to be smart enough to mark the branch in the upload process. This way we can see if coverage degrades in a PR

bobby-brennan: Here it is running currently: https://codecov.io/gh/reactiveops/polaris/commits

robscott: nice, that's awesome!

PR #92, opened 2019-05-16
Improving error handling

robscott:
There were a number of places where errors were not being logged or displayed, this should fix a bunch of them.

### DIFF
--- a/main.go
+++ b/main.go
@@ -89,10 +89,18 @@ func main() {
 	if *webhook {
 		startWebhookServer(c, *disableWebhookConfigInstaller, *webhookPort)
 	} else if *dashboard {
-		k, _ := kube.CreateResourceProvider(*auditPath)
+		k, err := kube.CreateResourceProvider(*auditPath)
+		if err != nil {
+			logrus.Errorf("Error fetching Kubernetes resources %v", err)
+			os.Exit(1)
+		}
 		startDashboardServer(c, k, *dashboardPort)
 	} else if *audit {
-		k, _ := kube.CreateResourceProvider(*auditPath)
+		k, err := kube.CreateResourceProvider(*auditPath)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/kube/resources.go
```diff
@@ -91,12 +94,14 @@ func CreateResourceProviderFromPath(directory string) (*ResourceProvider, error)
...
 		contents, err := ioutil.ReadFile(path)
 		if err != nil {
+			logrus.Errorf("Error reading file %v", path)
 			return err
 		}
 		specs := regexp.MustCompile("\n-+\n").Split(string(contents), -1)
 		for _, spec := range specs {
 			err = addYaml(spec)
 			if err != nil {
+				logrus.Errorf("Error parsing YAML %v", err)
```
bobby-brennan: FYI this error will be logged twice (once in `addYaml`), but maybe for the best

main.go
```diff
@@ -89,10 +89,18 @@ func main() {
 	if *webhook {
 		startWebhookServer(c, *disableWebhookConfigInstaller, *webhookPort)
 	} else if *dashboard {
-		k, _ := kube.CreateResourceProvider(*auditPath)
+		k, err := kube.CreateResourceProvider(*auditPath)
```
bobby-brennan: These will get logged twice as well.

pkg/kube/resources.go
robscott: Yeah, that was intentional. I wanted to have more specific logging in `addYaml` but also add this as a failsafe if an error hadn't already gotten logged there. It will likely be repetitive if they run into an error, but does ensure we catch things added to `addYaml` in the future that may not get the same kind of logging in the future. I don't know, that could be overkill.

bobby-brennan: 👍 sounds good to me

Issue #94, opened 2019-05-16
Running polaris locally complains about dashboard.gohtml


nskoufis:
**Steps to reproduce**:

1. Install `polaris` on my local machine with brew
2. Download the default `config-full.yaml` from the root of this repo, change the name to `config.yaml` and make sure I'm in the same directory
3. Auth to my k8s cluster
4. Run `polaris --dashboard`. Log output
```
INFO[0009] Starting Polaris dashboard server on port 8080
```
5. Navigate to `localhost:8080`

**Expected behaviour**:
I see some sort of dashboard

**Actual behaviour**:
Console logs show
```
INFO[0021] Error getting template data stat /Users/<my username>/Code/polaris/templates/dashboard.gohtml: no such file or directory
```
Browser shows `Error getting template data`

robscott:
Hey @nskoufis, thanks for reporting this! The binary release was the thing we added most recently, and apparently we didn't test it thoroughly enough. We're using packr to include templates in binary releases, but it looks like that didn't get added to our goreleaser code for binary/homebrew releases. I'll try to get a fix out for this tomorrow.

bobby-brennan:
Fixed in `0.1.1`. Thanks for the report!

Issue #93, opened 2019-05-16
Include default config with distributed binaries


robscott:
When running the binary release, it will break when there isn't a config.yaml file present. We should find a way to include default config directly in the binary distribution.

bobby-brennan:
`packr` can help here

robscott:
#97 fixed this

PR #97, opened 2019-05-17
use packr for config.yaml

bobby-brennan:


bobby-brennan:
Had to move `config.yaml` to a separate directory, as packr needs a directory as input, and every file in that directory ends up in the binary (so root directory is a bad idea).

bobby-brennan:
Fixes #94 and #93

PR #127, opened 2019-05-23
Better support for CI/CD use case

bobby-brennan:
Goal here was better support for CI/CD, but it ended up being a bit of a rabbit hole.

A few notes:
* Changed YAML marshaler to one that formats consistently with JSON marshaler
* Added score to output
* Added `output-format` which can be `json`, `yaml`, or `score` - the last option there just prints the overall score, e.g. `93`
* Changes to docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,14 @@ Polaris helps keep your cluster healthy. It runs a variety of checks to ensure t
 
 **Want to learn more?** ReactiveOps holds [office hours on Zoom](https://zoom.us/j/951540265) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@reactiveops.com`
 
+## Quickstart
+
+```
+kubectl apply -f https://raw.githubusercontent.com/reactiveops/polaris/master/deploy/dashboard.yaml
+kubectl port-forward --namespace polaris svc/polaris-dashboard 8080:80
+```
+With the port forwarding in place, you can open http://localhost:8080 in your browser to view the dashboard.
+
 ## Dashboard
 
 The Polaris dashboard is a way to get a simple visual overview of the current state of your Kubernetes deployments as well as a roadmap for what can be improved. The dashboard provides a cluster wide overview as well as breaking out results by category, namespace, and deployment.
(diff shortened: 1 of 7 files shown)


Gopkg.toml
```diff
@@ -60,3 +60,7 @@
 [[constraint]]
   revision = "v1.3.1"
   name = "github.com/golang/protobuf"
+
+[[constraint]]
+  name = "github.com/ghodss/yaml"
```
robscott: Instead of adding another yaml package, we could likely use the already existing one built into k8s that supports both JSON and YAML. It's worked well for me in the past and would ensure that we're using the same parser/decoder as k8s itself: https://godoc.org/k8s.io/apimachinery/pkg/util/yaml

bobby-brennan: Hmm...from what I can tell that library doesn't seem to have a Marshaler

bobby-brennan: Fixed per discussion offline

README.md
```diff
@@ -21,6 +21,14 @@ Polaris helps keep your cluster healthy. It runs a variety of checks to ensure t
 
 **Want to learn more?** ReactiveOps holds [office hours on Zoom](https://zoom.us/j/951540265) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@reactiveops.com`
 
+## Quickstart
```
endzyme: There are documentation changes not noted in the PR change set (non-blocking comment).

README.md
```diff
@@ -31,63 +39,84 @@ The Polaris dashboard is a way to get a simple visual overview of the current st
...
 kubectl apply -f https://raw.githubusercontent.com/reactiveops/polaris/master/deploy/dashboard.yaml
+kubectl port-forward --namespace polaris svc/polaris-dashboard 8080:80
 ```
-
-Polaris can also be deployed with Helm:
-
+#### Webhook
 ```
-helm upgrade --install polaris deploy/helm/polaris/ --namespace polaris
+kubectl apply -f https://raw.githubusercontent.com/reactiveops/polaris/master/deploy/webhook.yaml
```
endzyme: Looks like some of these are due to a squash merge - I will see if I can rebase to resolve the visual inconsistency.

README.md
bobby-brennan: good catch, fixed!

pkg/validator/types.go
```diff
@@ -45,6 +45,12 @@ type CountSummary struct {
 	Errors    uint
 }
 
+// GetScore returns an overall score in [0, 100] for the CountSummary
+func (cs *CountSummary) GetScore() uint {
```
endzyme: Is there documentation around this weighting process? (I.e. warnings decrease your score 1/2 as much as errors)? (non-blocking)

bobby-brennan: It's mentioned on the dashboard itself - not sure if you'd call that "documentation" :)

endzyme: oh no worries - not a huge deal - :) thanks!

PR #136, opened 2019-06-05
add base path option to helm chart

bobby-brennan:


bobby-brennan:
Also see: https://github.com/reactiveops/charts/pull/89/files

bobby-brennan:
Helm chart has been moved to reactiveops/charts

PR #138, opened 2019-06-05
Show cluster name/host on dashboard

bobby-brennan:
Addresses https://github.com/reactiveops/polaris/issues/124

Uses whatever the user specifies as `--cluster-name`, falling back to the `host` named in kubeconfig. Doesn't look like we can get the `name` field: https://github.com/kubernetes/client-go/issues/530  

Any potential issues with surfacing the `host` in the dashboard? E.g. could it have basic auth creds?

Here's what it looks like:

Screen Shot 2019-06-05 at 4 53 43 PM

### DIFF
--- a/pkg/kube/resources_test.go
+++ b/pkg/kube/resources_test.go
@@ -4,6 +4,7 @@ import (
 	"github.com/reactiveops/polaris/test"
 	"github.com/stretchr/testify/assert"
 	"testing"
+	"time"
 )
 
 func TestGetResourcesFromPath(t *testing.T) {
@@ -14,6 +15,7 @@ func TestGetResourcesFromPath(t *testing.T) {
 	assert.Equal(t, "Path", resources.SourceType, "Should have type Path")
 	assert.Equal(t, "./test_files/test_1", resources.SourceName, "Should have filename as name")
 	assert.Equal(t, "unknown", resources.ServerVersion, "Server version should be unknown")
+	assert.IsType(t, time.Now(), resources.CreationTime, "Creation time should be set")
 
 	assert.Equal(t, 0, len(resources.Nodes), "Should not have any nodes")
 
@@ -36,6 +38,7 @@ func TestGetMultipleResourceFromSingleFile(t *testing.T) {
 	assert.Equal(t, "Path", resources.SourceType, "Should have type Path")
 	assert.Equal(t, "./test_files/test_2/multi.yaml", resources.SourceName, "Should have filename as name")
 	assert.Equal(t, "unknown", resources.ServerVersion, "Server version should be unknown")
+	assert.IsType(t, time.Now(), resources.CreationTime, "Creation time should be set")
 
 	assert.Equal(t, 0, len(resources.Nodes), "Should not have any nodes")
 
@@ -55,6 +58,7 @@ func TestGetResourceFromAPI(t *testing.T) {
 
 	assert.Equal(t, "Cluster", resources.SourceType, "Should have type Path")
 	assert.Equal(t, "test", resources.SourceName, "Should have source name")
+	assert.IsType(t, time.Now(), resources.CreationTime, "Creation time should be set")
 
 	assert.Equal(t, 0, len(resources.Nodes), "Should not have any nodes")
 	assert.Equal(t, 1, len(resources.Deployments), "Should have a deployment")
--- a/pkg/kube/resources_test.go
+++ b/pkg/kube/resources_test.go
@@ -11,6 +11,8 @@ func TestGetResourcesFromPath(t *testing.T) {
 
 	assert.Equal(t, nil, err, "Error should be nil")
 
+	assert.Equal(t, "Path", resources.SourceType, "Should have type Path")
+	assert.Equal(t, "./test_files/test_1", resources.SourceName, "Should have filename as name")
 	assert.Equal(t, "unknown", resources.ServerVersion, "Server version should be unknown")
 
 	assert.Equal(t, 0, len(resources.Nodes), "Should not have any nodes")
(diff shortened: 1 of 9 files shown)


bobby-brennan:
Hmm that's a good point on `cluster-name`. The idea here was just to have something in the GUI to help users with many clusters know which report they were looking at. I changed the name of the flag to `display-name` to avoid confusion.

This also led me down a bit of a rabbit hole - I realized we were calling everything "cluster" even though we can run on local files as well. I added `SourceType` (cluster/path) and `SourceName` (host or pathname) to the audit output to reflect this.

As an aside, I think the right way to switch between which cluster you're working on is to use `kubectl config use-context`

What browser/OS are you using where the text is cut off? I'm not able to reproduce, but I see it in your screenshot.

kimschles:
👍 I think `display-name` is more descriptive of what the user is hoping to see. 

I'm using google chrome on Mac OS to test this branch.

pkg/validator/fullaudit.go
```diff
@@ -1,13 +1,15 @@
...
+	"time"
+
 	conf "github.com/reactiveops/polaris/pkg/config"
 	"github.com/reactiveops/polaris/pkg/kube"
 )
 
 const (
 	// PolarisOutputVersion is the version of the current output structure
-	PolarisOutputVersion = "0.0"
+	PolarisOutputVersion = "0.1"
```
endzyme: Will this versioning conflict be an issue? https://github.com/reactiveops/polaris/pull/127/files#diff-d344df2f44128868fd3b47e742d46d93R10 

Another PR is already planning to bump this from 0.0 to 0.1 (but the OutpufVersion will technically be different after this PR merges)

pkg/kube/resources_test.go
```diff
@@ -11,6 +11,8 @@ func TestGetResourcesFromPath(t *testing.T) {
 
 	assert.Equal(t, nil, err, "Error should be nil")
 
+	assert.Equal(t, "Path", resources.SourceType, "Should have type Path")
+	assert.Equal(t, "./test_files/test_1", resources.SourceName, "Should have filename as name")
```
endzyme: Is there value in also testing that `CreationTime` attribute exists, or is that not super important in this context?

endzyme: Nevermind - the CreationTime is added via the `CreateResourceProviderFromAPI` (moving comment down)

pkg/kube/resources_test.go
```diff
@@ -46,9 +50,12 @@ func TestGetMultipleResourceFromSingleFile(t *testing.T) {
 func TestGetResourceFromAPI(t *testing.T) {
 	k8s := test.SetupTestAPI()
 	k8s = test.SetupAddDeploys(k8s, "test")
-	resources, err := CreateResourceProviderFromAPI(k8s)
+	resources, err := CreateResourceProviderFromAPI(k8s, "test")
```
endzyme: Is there value in also testing that CreationTime attribute exists, or is that not super important in this context?

endzyme:
RE: The question of showing the cluster host (from kubectl) in the audit results; I think that's pretty safe from my perspective. A report that shows both the "config flaws" and the "http endpoint" of the cluster itself could be seen as a risk, but I think generally people will want to keep their reports private anyway.

Otherwise this change looks great, my only outstanding question revolves around testing the presence of a timestamp field as well a potential version conflict with another PR.

pkg/kube/resources_test.go
bobby-brennan: Good call!

pkg/validator/fullaudit.go
bobby-brennan: Yup, good eyes!

bobby-brennan:
@kimschles was able to repro the padding issue on chrome - should be fixed now!

bobby-brennan:
@kimschles I think I need your explicit approval (files changed -> review changes -> approve) since you requested changes. 

Thanks all for the valuable feedback!

kimschles:
I was also able to run the dashboard locally and I saw the changes you made. 🎉

alexcurtin:
@bobby-brennan how do you use this? I tried values.dashboard.clusterName and then the dashboard pod won't spin up due to: 
```flag provided but not defined: -cluster-name
```

endzyme:
I think with the latest version of Polaris were leveraging the helm chart located at [helm hub](https://hub.helm.sh/charts/reactiveops-stable/polaris) or [via GitHub](https://github.com/reactiveops/charts/blob/master/stable/polaris/README.md). 

I think the configuration can be provided via the config map yaml by setting it as a top level config `displayName: ...`. 

I've opened issue #158 to note the lack of docs for setting it via YAML. 

Feel free to hit us up in the issue if you have more questions. Thanks!!

alexcurtin:
@endzyme hmm, I tried many different ways, including adding it in the config and it now sayings flag provided but not defined: -display-name

```apiVersion: v1
data:
  config.yaml: |
    displayName: hello-world
    resources:
      cpuRequestsMissing: warning
      cpuLimitsMissing: warning
      memoryRequestsMissing: warning
      memoryLimitsMissing: warning
    images:
      tagNotSpecified: error
    healthChecks:
      readinessProbeMissing: warning
      livenessProbeMissing: warning
    networking:
      hostNetworkSet: warning
      hostPortSet: warning
    security:
      hostIPCSet: error
      hostPIDSet: error
      notReadOnlyRootFileSystem: warning
      privilegeEscalationAllowed: error
      runAsRootAllowed: warning
      runAsPrivileged: error
      capabilities:
        error:
          ifAnyAdded:
            - SYS_ADMIN
            - NET_ADMIN
            - ALL
        warning:
          ifAnyAddedBeyond:
            - CHOWN
            - DAC_OVERRIDE
            - FSETID
            - FOWNER
            - MKNOD
            - NET_RAW
            - SETGID
            - SETUID
            - SETFCAP
            - SETPCAP
            - NET_BIND_SERVICE
            - SYS_CHROOT
            - KILL
            - AUDIT_WRITE
kind: ConfigMap```

alexcurtin:
Also the stable chart is 1.5 so doesn't include that feature afaict.

Issue #141, opened 2019-06-08
Adjust potential os.exit(1) for resource creation


endzyme:
There is a call to `config.GetConfigOrDie()` in [resource.go:123](https://github.com/reactiveops/polaris/blob/12e10f911160f1234cc22a69a10e8cd25110aa70/pkg/kube/resources.go#L123). This attempts to load the config, but if unsuccessful will call an `os.Exit(1)` from inside `resourse.go`. 

From a user experience perspective it may be better to leverage `config.GetConfig()` and handle the `err != nil` by returning an empty `ResourceProvider` and an error message. Where this function is invoked, [already has error handling](https://github.com/reactiveops/polaris/blob/eab320af381a82c30694fa578d456f6df74facf5/main.go#L181) which would capture a more usable message to the user instead of calling an `os.Exit(1)`.

PR #144, opened 2019-06-11
adjust os exit

kimschles:
Fixes issue [#141](https://github.com/reactiveops/polaris/issues/141)

**What this PR does and why we need it**

I updated the `CreateResourceProviderFromCluster` function to find the KubeConfig and return an error if there is an issue retrieving the data. I wasn't able to figure out how to return the empty `ResourceProvider`, so I returned `nil` instead. Let me know if the empty `ResourceProvider` is necessary and I'll figure that out. 

See line 132 for the specific line I'm referring to: https://github.com/reactiveops/polaris/compare/ks/adjust-os-exit?expand=1#diff-5524e4ddda8206adfe02f6a98ed7160eR132

### DIFF
--- a/pkg/kube/resources.go
+++ b/pkg/kube/resources.go
@@ -126,7 +126,11 @@ func CreateResourceProviderFromPath(directory string) (*ResourceProvider, error)
 
 // CreateResourceProviderFromCluster creates a new ResourceProvider using live data from a cluster
 func CreateResourceProviderFromCluster() (*ResourceProvider, error) {
-	kubeConf := config.GetConfig()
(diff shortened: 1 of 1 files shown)


bobby-brennan:
Looks great, thanks Kim!

The caller will likely just bail out if an error is returned (as currently happens in main.go and dashboard.go), so no need for an empty ResourceProvider.

bobby-brennan:
Also, for future reference, I think you could do something like `return &ResourceProvider{}`. But it's probably better to return null in this context.

Issue #146, opened 2019-06-12
Issue with runAsNonRoot


jclarksnps:
Not entirely sure that your check for runAsNonRoot is working, or we mis-understand exactly what it's checking.  We have a pod running which is set at the pod level as below, yet reactive is still saying it shouldn't run as root... which it isn't.  Since the securityContext at the container level is only for overriding what is set at the pod level I hope that being set at the pod level is enough.  Any ideas?

securityContext:
        runAsNonRoot: true
        runAsUser: 5000

endzyme:
Thanks @jclarksnps - after looking into this, it does look to be a bug. I have been able reproduce locally. I think the missing context here is that the securityContext is not set on the Container but it is set at the pod level. This is innate kube logic which is missing from `polaris`. I can file this as a bug and we'll look at fixing this up in the validation logic. Thanks for the note!!

jclarksnps:
Thanks!

endzyme:
#149 Linking

jclarksnps:
Thanks for fixing this, any eta on a release?

endzyme:
Hi @jclarksnps - I'll get back to you shortly on the timing :smiley: - Thanks again!

endzyme:
@jclarksnps - I am planning on cutting a release tomorrow if that works for you! :)

jclarksnps:
Thanks! Yep, we'll deploy and check it out.

endzyme:
https://github.com/reactiveops/polaris/releases/tag/0.2.1 -- all set 

thanks for your patience!

PR #149, opened 2019-06-13
#146 Fixing Container Security Context Logic

endzyme:
Kubernetes rationalizes Container Security Context in conjuction with the
Pod Spec Security Context. In this scenario you can 'leave out' certain
security context settings and rely on the pod spec definition to still
set these settings for you. The RunAsNonRoot setting originally only checked
to see if the value was set at the container level, vs also checking if it
was enabled at the pod level.

I have attached the container's parent pod spec to the container validate
struct in case any other things like this arise in the future.

I have also refactored the logic for validating bool pointers, since these
can be tricky, if you want to avoid dereferences pointer issues.

Changes:
- Added parent pod spec of container to validate certain settings which affect container spec
- Refactored the logic statements for validating bool pointers (used helpers)
- Added tests for this pod.container.securityContext condition


endzyme:
I'm interested if anyone has any other ideas on how to solve this issue of podSpec affecting unset values in containers. Another potential way to go about this is to check each container (while in the pod validation) if the security-context is set to `false || nil` at the pod level.

endzyme:
FYI I'm still missing the CHANGELOG additions (when this eventually is ready)

pkg/validator/bool_pointer_helpers.go
```diff
@@ -0,0 +1,55 @@
+package validator
+
```
bobby-brennan: Instead of all this, I think we can get away with a single function:

```go
func boolValue(boolPtr *bool, default bool) bool {
  if boolPtr == nil {
    return default
  }
  return *boolPtr
}
```

(note: you can do `nil` comparison without casting to a bool - I know we were casting before but I don't think that buys us anything.)

Then you can do e.g. `getBoolVal(securityContext.RunAsNonRoot, getBoolVal(podSecurityContext.RunAsNonRoot, false))`

Can probably just put it in `container.go` if it works.

pkg/validator/pod.go
```diff
@@ -54,7 +55,7 @@ func ValidatePod(podConf conf.Configuration, pod *corev1.PodSpec) PodResult {
 
 func (pv *PodValidation) validateContainers(containers []corev1.Container, pRes *PodResult, podConf *conf.Configuration, isInit bool) {
 	for _, container := range containers {
-		cRes := ValidateContainer(podConf, &container, isInit)
+		cRes := ValidateContainer(podConf, &container, isInit, pRes)
```
bobby-brennan: While you're in here, probably makes sense to match the parameter order with the above, i.e. `(&container, pres, podConf, isInit)`

bobby-brennan:
This is awesome, thanks for chasing it down! Tests look solid.

I looked a bit at [SecurityContext](https://godoc.org/k8s.io/api/core/v1#SecurityContext) vs [PodSecurityContext](https://godoc.org/k8s.io/api/core/v1#PodSecurityContext), and this seems to be the only variable that's shared between the two.

Re: changelog, I've just been updating at release time (since we don't know the eventual version number). You're welcome to start an "In Progress" section though.

Other than the comment on bool helpers this LGTM.

pkg/validator/container.go
```diff
@@ -176,39 +194,55 @@ func (cv *ContainerValidation) validateNetworking(networkConf *conf.Networking)
...
+	// Support an empty pod security context
+	if podSecurityContext == nil {
+		podSecurityContext = &corev1.PodSecurityContext{}
+	}
+
 	if securityConf.RunAsRootAllowed.IsActionable() {
-		if securityContext.RunAsNonRoot == (*bool)(nil) || !*securityContext.RunAsNonRoot {
-			cv.addFailure(messages.RunAsRootFailure, securityConf.RunAsRootAllowed, category)
-		} else {
+		if getBoolValue(securityContext.RunAsNonRoot) {
```
bobby-brennan: Non-blocking: I think `getBoolValue(securityContext.RunAsNonRoot) || getBoolValue(podSecurityContext.RunAsNonRoot)` would simplify this a bit

pkg/validator/container.go
```diff
@@ -323,3 +357,12 @@ func capContains(list []corev1.Capability, val corev1.Capability) bool {
 
 	return false
 }
+
+// getBoolValue returns false if nil or returns the value of the bool pointer
+func getBoolValue(val *bool) bool {
```
bobby-brennan: Even simpler!

pkg/validator/container.go
endzyme: I found one more case where my code and this `||` implementation will return incorrect results. 

If the podspec is set to True, but the Container spec is explicitly False, then Polaris will not find a problem (when one exists). 

I have updated the code (pushing now)

PR #155, opened 2019-06-24
Add exit code flag 

kimschles:
**What this PR does** 
This PR creates an `--set-exit-code-on-error` flag. If the user runs an audit and passes in this flag, the audit will exit with a status code of `3`. This functionality is most helpful for CI/CD use cases. 

**This PR addresses issue** [143](https://github.com/reactiveops/polaris/issues/143)

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,7 +50,7 @@ func main() {
 	webhook := flag.Bool("webhook", false, "Runs the webhook webserver.")
 	audit := flag.Bool("audit", false, "Runs a one-time audit.")
 	auditPath := flag.String("audit-path", "", "If specified, audits one or more YAML files instead of a cluster")
-	setExitCode := flag.Bool("exit-code", false, "set an exit code of 2 when the audit contains error-level issues.")
+	setExitCode := flag.Bool("set-exit-code-on-error", false, "set an exit code of 2 when the audit contains error-level issues.")
 	dashboardPort := flag.Int("dashboard-port", 8080, "Port for the dashboard webserver")
 	dashboardBasePath := flag.String("dashboard-base-path", "/", "Path on which the dashboard is served")
 	webhookPort := flag.Int("webhook-port", 9876, "Port for the webhook webserver")
@@ -194,8 +194,8 @@ func runAudit(c conf.Configuration, auditPath string, setExitCode bool, outputFi
 	}
 
 	if setExitCode && auditData.ClusterSummary.Results.Totals.Errors > 0 {
-		logrus.Errorf("Error found. Exiting audit.")
-		os.Exit(2)
+		logrus.Infof("Error found. Exiting audit.")
+		os.Exit(3)
 	}
 
 	var outputBytes []byte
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -50,6 +50,7 @@ func main() {
 	webhook := flag.Bool("webhook", false, "Runs the webhook webserver.")
 	audit := flag.Bool("audit", false, "Runs a one-time audit.")
 	auditPath := flag.String("audit-path", "", "If specified, audits one or more YAML files instead of a cluster")
+	setExitCode := flag.Bool("exit-code", false, "set an exit code of 2 when the audit contains error-level issues.")
```
bobby-brennan: Can you change from `exit-code` to `set-exit-code-on-error`?

main.go
```diff
@@ -192,6 +193,11 @@ func runAudit(c conf.Configuration, auditPath string, outputFile string, outputU
 		panic(err)
 	}
 
+	if setExitCode && auditData.ClusterSummary.Results.Totals.Errors > 0 {
+		logrus.Errorf("Error found. Exiting audit.")
```
bobby-brennan: I think this can be `Infof`, since it's expected.

main.go
```diff
@@ -192,6 +193,11 @@ func runAudit(c conf.Configuration, auditPath string, outputFile string, outputU
...
+		os.Exit(2)
```
bobby-brennan: Did some research - apparently `2` [has special meaning](http://www.tldp.org/LDP/abs/html/exitcodes.html)

Can you change to `3`?

kimschles:
@bobby-brennan I made the changes your requested.

main.go
```diff
@@ -192,6 +193,11 @@ func runAudit(c conf.Configuration, auditPath string, outputFile string, outputU
 		panic(err)
 	}
 
+	if setExitCode && auditData.ClusterSummary.Results.Totals.Errors > 0 {
```
bobby-brennan: Sorry, I should have caught this before - this block should go at the very bottom of the function. That way we'll still get to see the audit output.

kimschles: @bobby-brennan I moved that block to the bottom of the function. 😄

bobby-brennan:
Thanks Kim, great work!

main.go
```diff
@@ -50,6 +50,7 @@ func main() {
 	webhook := flag.Bool("webhook", false, "Runs the webhook webserver.")
 	audit := flag.Bool("audit", false, "Runs a one-time audit.")
 	auditPath := flag.String("audit-path", "", "If specified, audits one or more YAML files instead of a cluster")
+	setExitCode := flag.Bool("set-exit-code-on-error", false, "set an exit code of 2 when the audit contains error-level issues.")
```
endzyme: Heads up - this doc doesn't match the `os.Exit(3)` used below. Also, any consideration on just using `1` as the exit code?

bobby-brennan: Will fix.

We decided against `1` so that we could differentiate this expected failure from unexpected failures.

PR #156, opened 2019-06-25
add deploy configurations to source control

bobby-brennan:
Slight hiccup with this process: the new deploy.yaml will get checked in before the images are built. So there will be a 10-15 minute period of breakage.

Maybe it'd be better to tag the commit *before* it gets merged - that way images are built ahead of time. And we can run goreleaser only once that tag is merged into master.

Thoughts?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,8 +62,8 @@ Each new pull request should:
 Patch releases only need to change this repo. The Helm chart and deploy scripts
 will automatically pull in the latest changes.
 
-If the release involves changes to anything in the `deploy/` folder, it needs to be a minor
-or major release in order to prevent breaking the Helm chart.
+If the release involves changes to anything in the `deploy/` folder (e.g. new RBAC permissions),
+it needs to be a minor or major release in order to prevent breaking the Helm chart.
 
 To deploy a patch release, follow steps 2 and 3 below.
 
@@ -77,7 +77,8 @@ The steps are:
         1. stable/polaris/README.md
         2. stable/polaris/Chart.yaml
         3. stable/polaris/values.yaml
-    2. Make any additional changes necessary (e.g. new RBAC permissions)
+    2. Make any necessary changes to the chart to support the new version of Polaris (e.g. new RBAC permissions)
+    3. **Don't merge yet!**
 2. Create a PR for this repo
     1. Regenerate the deployment files. Assuming you've cloned the charts repo to `~/charts`:
         1. `helm template ~/charts/stable/polaris/ --name polaris --namespace polaris --set templateOnly=true > deploy/dashboard.yaml`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,33 +58,38 @@ Each new pull request should:
 
 ## Creating a new release
 
-### Minor/patch releases
-Minor and patch releases only need to change this repo. The Helm chart and deploy scripts
+### Patch releases
+Patch releases only need to change this repo. The Helm chart and deploy scripts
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -58,33 +58,38 @@ Each new pull request should:
...
-1. Create a PR in the [charts repo](https://github.com/reactiveops/charts/)
-    1. Use a branch named `polaris-latest`
-    2. Bump the version number in:
+1. Modify the [Helm chart](https://github.com/reactiveops/charts/stable/polaris)
+    1. Bump the version number in:
         1. stable/polaris/README.md
         2. stable/polaris/Chart.yaml
         3. stable/polaris/values.yaml
-    3. **Don't merge yet!**
+    2. Make any additional changes necessary (e.g. new RBAC permissions)
```
endzyme: This sentence is a little vague - is it specifically calling out the `./deploy` manifests that may need to be changed?

CONTRIBUTING.md
```diff
@@ -58,33 +58,38 @@ Each new pull request should:
...
+Patch releases only need to change this repo. The Helm chart and deploy scripts
 will automatically pull in the latest changes.
 
-To deploy a minor or patch release, follow steps 2 and 3 from "Major releases" below.
+If the release involves changes to anything in the `deploy/` folder, it needs to be a minor
+or major release in order to prevent breaking the Helm chart.
 
-### Major releases
-Major releases need to change both this repository and the
+To deploy a patch release, follow steps 2 and 3 below.
```
endzyme: How would you feel about kicking DRY to the curb, and completely separating the instructions for `patch` vs `minor/major`? My gut feel, in running through this, is jumping from section to section and picking only the "parts you need" could lead to confusion. 

If someone wants to  update instructions 2&3 later in life, it may be unclear that they need to stand alone for the patch release process. Food for thought and happy to discuss.

CONTRIBUTING.md
```diff
@@ -58,33 +58,38 @@ Each new pull request should:
...
         1. stable/polaris/README.md
         2. stable/polaris/Chart.yaml
         3. stable/polaris/values.yaml
-    3. **Don't merge yet!**
+    2. Make any additional changes necessary (e.g. new RBAC permissions)
 2. Create a PR for this repo
-    1. Bump the version number in:
+    1. Regenerate the deployment files. Assuming you've cloned the charts repo to `~/charts`:
+        1. `helm template ~/charts/stable/polaris/ --name polaris --namespace polaris --set templateOnly=true > deploy/dashboard.yaml`
+        2. `helm template ~/charts/stable/polaris/ --name polaris --namespace polaris --set templateOnly=true --set webhook.enable=true --set dashboard.enable=false > deploy/webhook.yaml`
```
endzyme: Since we're leveraging the chart for this, and the chart downloads polaris using a "latest minor" type version (0.2 at the time of writing this); This would introduce an unexpected behavior while the PR+Tag builds are running. Anyone who runs the `kubectl apply -f deploy.yml` before the images are pushed to quay, will accidentally download `${CURRENT_VERSION} - 1`. This could lead to very unexpected behavior for users while pipelines are in flight.

I think maybe a different approach could be to explicitly call out the specific version of code we're pushing as a part of the release. Using the same process, while the images are pushing to quay, users would get an `ImagePullBackOff` if they run the `kubectl apply -f deploy.yml` before the images are in quay.

bobby-brennan:
Note: the 10-15 minute breakage mentioned above has been solved. The README will continue to reference the deploy YAML on the releases page, which will always point to a valid Docker image.

PR #157, opened 2019-06-26
Set exit code if score drops below provided threshold

bobby-brennan:


README.md
```diff
@@ -115,14 +118,12 @@ polaris --audit --audit-path ./deploy/
 
 ##### Running with CI/CD
 You can integrate Polaris into CI/CD for repositories containing infrastructure-as-code.
-For example, to fail whenever the Polaris score drops below 90%:
+For example, to fail whenever the Polaris score drops below 90%, or Polaris detects error-level
```
endzyme: I think this maybe should call out errors first - since it's prioritized in the if logic. I, personally, would also call out:
>"..to fail if Polaris detects _any_ error-level issues, or if there are no error-level issues, fail if the score drops below 90%"

main.go
```diff
@@ -96,7 +97,14 @@ func main() {
 	} else if *dashboard {
 		startDashboardServer(c, *auditPath, *dashboardPort, *dashboardBasePath)
 	} else if *audit {
-		runAudit(c, *auditPath, *setExitCode, *auditOutputFile, *auditOutputURL, *auditOutputFormat)
+		auditData := runAudit(c, *auditPath, *auditOutputFile, *auditOutputURL, *auditOutputFormat)
+		if *setExitCode && auditData.ClusterSummary.Results.Totals.Errors > 0 {
+			logrus.Infof("%d errors found in audit", auditData.ClusterSummary.Results.Totals.Errors)
```
endzyme: I think it might be useful for the end user to know the score at this point too so they can adjust the flags armed with some knowledge if they're playing with what's important to them.

I'm happy to chat about it if you want to consider it.

main.go
```diff
@@ -50,7 +50,8 @@ func main() {
 	webhook := flag.Bool("webhook", false, "Runs the webhook webserver.")
 	audit := flag.Bool("audit", false, "Runs a one-time audit.")
 	auditPath := flag.String("audit-path", "", "If specified, audits one or more YAML files instead of a cluster")
-	setExitCode := flag.Bool("set-exit-code-on-error", false, "set an exit code of 2 when the audit contains error-level issues.")
+	setExitCode := flag.Bool("set-exit-code-on-error", false, "When running with --audit, set an exit code of 3 when the audit contains error-level issues.")
+	minScore := flag.Int("set-exit-code-below-score", 0, "When running with --audit, set an exit code of 4 when the score is below this threshold (1-100)")
```
endzyme: I don't think it's overly clear that these actually change the output behavior of the audit run when you set these. My expected behavior would be that these only `--set-exit-code-[on-condition]` vs changing the audit reports _and_ change the exit code. Happy to discuss more

main.go
endzyme: So still reading through all this and looks like `runAudit()` still outputs the report to stdout and the flags just add a note to stderr of the errors found and if the reported score is lower than the desired minimum 

:+1: 

I was confused because auditData := runAudit() has a side-effect of printing to stdout which wasn't immediately clear to me since we're storing "outputs" to a variable.

main.go
```diff
@@ -254,9 +262,5 @@ func runAudit(c conf.Configuration, auditPath string, setExitCode bool, outputFi
 			}
 		}
 	}
-
```
endzyme: this is a good move to the other main function :+1:

main.go
```diff
@@ -181,7 +189,7 @@ func startWebhookServer(c conf.Configuration, disableWebhookConfigInstaller bool
 	}
 }
 
-func runAudit(c conf.Configuration, auditPath string, setExitCode bool, outputFile string, outputURL string, outputFormat string) {
+func runAudit(c conf.Configuration, auditPath string, outputFile string, outputURL string, outputFormat string) validator.AuditData {
```
endzyme: nit pick - feels like the validator.AuditData class should hold the `marshal` functionality of the auditdata itself, instead of in `runAudit()`.

main.go
```diff
@@ -96,7 +97,14 @@ func main() {
...
 		startDashboardServer(c, *auditPath, *dashboardPort, *dashboardBasePath)
 	} else if *audit {
-		runAudit(c, *auditPath, *setExitCode, *auditOutputFile, *auditOutputURL, *auditOutputFormat)
+		auditData := runAudit(c, *auditPath, *auditOutputFile, *auditOutputURL, *auditOutputFormat)
+		if *setExitCode && auditData.ClusterSummary.Results.Totals.Errors > 0 {
+			logrus.Infof("%d errors found in audit", auditData.ClusterSummary.Results.Totals.Errors)
+			os.Exit(3)
+		} else if *minScore != 0 && auditData.ClusterSummary.Score < uint(*minScore) {
+			logrus.Infof("Audit score of %d is less than the provided minimum of %d", auditData.ClusterSummary.Score, *minScore)
+			os.Exit(4)
```
endzyme: Is there any way to write tests for this behavior for future refactors?

main.go
endzyme: ran through the code more in depth: https://github.com/reactiveops/polaris/pull/157#discussion_r297996606 -- resolving comment

main.go
bobby-brennan: Good point - renamed to `runAndReportAudit`

main.go
bobby-brennan: Kim and I thought about this - it's definitely possible, but non-trivial. It'd also be something stuffed into CircleCI rather than a golang test.

main.go
bobby-brennan: Not sure I agree here - marshaling is a pretty trivial task, and one I'd generally leave up to the user. Could be convinced though.

README.md
bobby-brennan: 👍

main.go
endzyme: In my mind it felt like a part of domain of the audit data. I just thought it would be nice to use methods like `func (ad AuditData) toJSON() []bytes` and `func (ad AuditData) toYAML() []bytes`.

PR #160, opened 2019-07-10
Rename org in GitHub API call

bobby-brennan:


bobby-brennan:
Yup, that's correct! Shooting for later this afternoon.

PR #162, opened 2019-07-10
update to 0.3.0

bobby-brennan:


bobby-brennan:
Good catch! I added all the other flags that were missing as well :)

Issue #107, opened 2019-05-17
Validate more resources than deployments


robscott:
As referenced in our [roadmap](https://github.com/reactiveops/polaris/blob/master/ROADMAP.md#new-controller-types), we want to add support for a variety of additional resources, starting with those that act as parent resources such as pods, including:

- StatefulSet
- DaemonSet
- Job
- CronJob
- ReplicaSet

marcecaro:
What about ReplicaControllers? I know that it is getting replaced by deployments, but it is still very used.

endzyme:
@marcecaro sure thing - I can add that to #166

bobby-brennan:
Implemented in #166

PR #166, opened 2019-07-16
Additional Pod Controller Scans

endzyme:
- Refactored the way controllers work to be an interface
- Added configurable controllers to include in scans
- Added daemonsets, jobs and cronjobs in scans

Explicitly did not add ReplicationControllers and ReplicaSets since they are either deprecated or controlled by deployments (generally).

- Added tests to supportedController scanner
- Added documentation on the exit codes 
- Added code documentation for longer functions


README.md
```diff
@@ -116,6 +116,20 @@ rather than a cluster:
 polaris --audit --audit-path ./deploy/
 ```
 
+##### Exit Codes for Audits
+<dl>
+  <dt>Exit 0</dt>
+  <dd>Successful exit code</dd>
+  <dt>Exit 1</dt>
+  <dd>Could not run audit, or application had a failure while running.</dd>
```
bobby-brennan: This might be better described as "unexpected error". We seem to use it like an HTTP `500`

README.md
```diff
@@ -116,6 +116,20 @@ rather than a cluster:
 polaris --audit --audit-path ./deploy/
 ```
 
+##### Exit Codes for Audits
```
bobby-brennan: The README is getting a little dense. What do you think about breaking this and the CLI flags into a separate file?

pkg/config/supportedcontrollers.go
```diff
@@ -0,0 +1,105 @@
...
+// SupportedController is a constant item of a controller that is supported for scanning pod specs
+type SupportedController int
+
+// String returns the string name for a given SupportedController enum
+func (s SupportedController) String() string {
+	return ControllerStrings[s]
+}
+
+// MarshalJSON manages writing the enum into json data or error on unsupported value
+func (s SupportedController) MarshalJSON() ([]byte, error) {
```
bobby-brennan: The Marshal and Unmarshal functions don't seem to be used...

pkg/config/supportedcontrollers.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+// GetSupportedControllerFromString fuzzy matches a string with a SupportedController Enum
+func GetSupportedControllerFromString(str string) (SupportedController, error) {
+	lowerStr := strings.ToLower(str)
+	if stringLookupForSupportedControllers[lowerStr] == Unsupported {
+		return 0, fmt.Errorf("Value ('%v') in configuration was not found in Supported Controllers: (%v)", str, strings.Join(ControllerStrings, ","))
+	}
+	return stringLookupForSupportedControllers[lowerStr], nil
```
bobby-brennan: Do we need to check if `lowerStr` is in the map?

pkg/webhook/validator.go
```diff
@@ -92,16 +93,27 @@ func (v *Validator) Handle(ctx context.Context, req types.Request) types.Respons
...
 		podResult = validator.ValidatePod(v.Config, &pod.Spec)
 	} else {
-		var controller validator.Controller
-		switch req.AdmissionRequest.Kind.Kind {
-		case "Deployment":
+		var controller controllers.Interface
+		if yes := v.Config.CheckIfKindIsConfiguredForValidation(req.AdmissionRequest.Kind.Kind); !yes {
+			logrus.Warnf("Skipping, kind (%s) isn't something we are configured to scan", req.AdmissionRequest.Kind.Kind)
+			// FIXME: Should we be returning an OK response as skipped?
+			return admission.ErrorResponse(http.StatusBadRequest, err)
```
bobby-brennan: Yes, I think for the webhook this should be an `OK` response for unsupported types. Otherwise (IIUC) if a new controller type is introduced, we'll reject it until Polaris gets updated to support it.

pkg/config/supportedcontrollers.go
endzyme: Hmm yes it would appear so. I'll add some tests around this and see what the default behavior is for nil retirns.

bobby-brennan:
You'll need to update the RBAC definitions in the chart to regenerate everything in the `./deploy` folder as well.

endzyme:
> You'll need to update the RBAC definitions in the chart to regenerate everything in the `./deploy` folder as well.

For sure, I saw that on your PR for the limit ranges. This is a WIP and I wanted to validate this isn't too over the top or out of line with the requested functionality.

bobby-brennan:
FYI, in https://github.com/FairwindsOps/polaris/issues/107 someone mentions that `ReplicationControllers` are still widely in use - maybe it makes sense to add them?

endzyme:
@bobby-brennan: I've incorporated the following (per request)

- Added `replicationController` type
- Added more testing (better coverage)
- Fixed webhook integration and support for skipping non-scanned controllers and other kinds

pkg/config/supportedcontrollers.go
endzyme: So this is a weird language behavior - i find it strange but it's supposedly supported :man_shrugging: 

https://blog.golang.org/go-maps-in-action

TL;DR; 
if you have a `map[string]int` and you call a key that doesn't exist - you get the `zero value` default of the _value_ type. So in our case `map[string]SupportedController` -- `SupportedController is an enum which has a value of `int`, and in this case `int`s "`zero value`" is 0. (for strings it's `""` and for `interfaces`, `func`, `struct`, you can get `nil` as the zero value)

README.md
```diff
@@ -22,16 +22,19 @@ problems in the future. Polaris can be run in a few different modes:
 - An experimental validating webhook that can prevent any future deployments that do not live up to a configured standard.
 - A command-line audit that can be incorporated into your CI/CD pipeline
 
-**Want to learn more?** ReactiveOps holds [office hours on Zoom](https://zoom.us/j/242508205) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@fairwinds.com`
+**Want to learn more?** FairWinds holds [office hours on Zoom](https://zoom.us/j/242508205) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@fairwinds.com`
```
bobby-brennan: `s/W/w/`

pkg/kube/resources.go
```diff
@@ -173,6 +199,18 @@ func addResourceFromString(contents string, resources *ResourceProvider) error {
 		dep := appsv1.StatefulSet{}
 		err = decoder.Decode(&dep)
 		resources.StatefulSets = append(resources.StatefulSets, dep)
+	} else if resource.Kind == "DaemonSet" {
+		dep := appsv1.DaemonSet{}
```
bobby-brennan: nit: I clearly copy/pasta'd the code for StatefulSet - `dep` was originally meant to refer to `Deployment`. Maybe change `dep` to `ctrl` everywhere?

pkg/webhook/validator.go
bobby-brennan: Still seeing the `FIXME` here, and looks like it's still `StatusBadRequest`.

bobby-brennan:
Looking great Nick! A few comments but I think this is just about ready.

I really like using the interface for different controller types, makes a lot more sense than what I had.

As a side note, we should start testing the webhook for correctness and feature parity. Opened https://github.com/FairwindsOps/polaris/issues/178 to track.

pkg/kube/resources.go
endzyme: Ah sounds good. Yeah I guess dep didn't connect for me :) I'll fix that up

pkg/webhook/validator.go
endzyme: I think maybe it was a bad browser cache on your side? I am not seeing it

README.md
```diff
@@ -22,16 +22,19 @@ problems in the future. Polaris can be run in a few different modes:
 - An experimental validating webhook that can prevent any future deployments that do not live up to a configured standard.
 - A command-line audit that can be incorporated into your CI/CD pipeline
 
-**Want to learn more?** ReactiveOps holds [office hours on Zoom](https://zoom.us/j/242508205) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@fairwinds.com`
+**Want to learn more?** Fairwinds holds [office hours on Zoom](https://zoom.us/j/242508205) the first Friday of every month, at 12pm Eastern. You can also reach out via email at `opensource@fairwinds.com`
 
-## Quickstart
+# Quickstart Dashboard
```
bobby-brennan: Can we rename this to `Dashboard Quickstart`?

pkg/dashboard/helpers.go
```diff
@@ -16,14 +16,14 @@ package dashboard
...
 import (
 	"fmt"
-	"github.com/fairwindsops/polaris/pkg/validator"
 	"strings"
+
+	"github.com/fairwindsops/polaris/pkg/validator"
 )
 
 func getAllControllerResults(nr validator.NamespaceResult) []validator.ControllerResult {
 	results := []validator.ControllerResult{}
```
bobby-brennan: You can drop this line and add change to `:=` below

PR #177, opened 2019-07-24
Pod tests

willfairwinds:
Added pod tests for configured Host IPC, Host PID, and Host Newtowrk, also fixed order of expected and actual results in the original test and all others.


pkg/validator/pod_test.go
```diff
@@ -71,6 +71,169 @@ func TestValidatePod(t *testing.T) {
...
+
+	expectedMessages := []*ResultMessage{
+		{Message: "Host network should not be configured", Type: "warning", Category: "Networking"},
+		{Message: "Host IPC is not configured", Type: "success", Category: "Security"},
+		{Message: "Host PID is not configured", Type: "success", Category: "Security"},
+	}
+
+	actualPodResult := ValidatePod(c, &pod.Spec)
+
+	assert.Equal(t, len(actualPodResult.ContainerResults), 1, "should be equal")
```
bobby-brennan: Expected and actual are flipped here (`1` is the expected value). Same above and below

endzyme:
Thanks for the submission @willfairwinds! We'll take a look this week but so far seems pretty straight forward.

bobby-brennan:
🎉  Looks good to me!

CLA wants you to sign again for some reason, but once that's done you're good to merge. Thanks again!

PR #181, opened 2019-07-25
Wl/config url

willfairwinds:
Expanded ParseFile to include URL's and added a test for it.


pkg/config/config.go
```diff
@@ -112,16 +114,27 @@ type SecurityCapabilityLists struct {
...
+			return Configuration{}, err
+		}
+		rawBytes, err = ioutil.ReadAll(response.Body)
 	} else {
-		rawBytes, err = ioutil.ReadFile(path)
-	}
-	if err != nil {
-		return Configuration{}, err
+		//path is local
+		configBox := packr.New("Config", "../../examples")
```
bobby-brennan: wasn't like this before, but you can put this underneath `if path == ""` to be a little cleaner and more efficient.

pkg/config/config.go
```diff
@@ -112,16 +114,27 @@ type SecurityCapabilityLists struct {
...
+		if path == "" {
+			rawBytes, err = configBox.Find("config.yaml")
+		} else {
+			rawBytes, err = ioutil.ReadFile(path)
+		}
+		if err != nil {
```
bobby-brennan: This needs to go outside the bigger `if` block, to handle the case where `ioutil.ReadAll` errors out

bobby-brennan:
Looks great! Just one change re: error handling and I think this is good to merge.

Bonus: you can try adding another test that _doesn't_ start an HTTP server (so that the HTTP request fails), and check to make sure you get the right error message. That would catch the current bug.

pkg/config/config_test.go
```diff
@@ -120,6 +124,35 @@ func TestParseJson(t *testing.T) {
...
+		panic(err)
+	}
+	testParsedConfig(t, &parsedConf)
+
+}
+
+func TestConfigNoServerError(t *testing.T) {
+	var err error
+	_, err = ParseFile("http://localhost:8081/exampleURL")
+	assert.EqualError(t, err, "Get http://localhost:8081/exampleURL: dial tcp [::1]:8081: connect: connection refused")
```
bobby-brennan: Looks like the message can change a bit based on environment (which is why CircleCI is failing). To fix the tests on CircleCI, you could change to something like:

```go
assert.Error(t, err)
assert.Regexp(t, regexp.MustCompile("connection refused"), err.Error())
```

pkg/config/config.go
```diff
@@ -112,16 +114,28 @@ type SecurityCapabilityLists struct {
 
 // ParseFile parses config from a file.
 func ParseFile(path string) (Configuration, error) {
-	configBox := packr.New("Config", "../../examples")
-	var rawBytes []byte
 	var err error
+	var rawBytes []byte
+	configBox := packr.New("Config", "../../examples")
 	if path == "" {
```
bobby-brennan: This won't work as-is. If path is `""`, it will still hit the `else` block below, and overwrite `rawBytes`. You'll see it fail if you run `go run main.go --dashboard --dashboard-port 3000 --audit-path ./examples`

If you change the `if strings.HasPrefix` line to an `else if`, that should fix the issue.

pkg/config/config.go
```diff
@@ -112,16 +114,27 @@ type SecurityCapabilityLists struct {
...
+			return Configuration{}, err
+		}
+		rawBytes, err = ioutil.ReadAll(response.Body)
 	} else {
+		//path is local
 		rawBytes, err = ioutil.ReadFile(path)
-	}
-	if err != nil {
-		return Configuration{}, err
+		if err != nil {
```
bobby-brennan: This still has to go outside the `else` statement, or the `err` on line 131 won't get caught

pkg/config/config.go
```diff
@@ -112,16 +114,27 @@ type SecurityCapabilityLists struct {
...
 // ParseFile parses config from a file.
 func ParseFile(path string) (Configuration, error) {
-	configBox := packr.New("Config", "../../examples")
-	var rawBytes []byte
 	var err error
+	var rawBytes []byte
+	configBox := packr.New("Config", "../../examples")
 	if path == "" {
 		rawBytes, err = configBox.Find("config.yaml")
+		if err != nil {
```
bobby-brennan: Once you move the `if err != nil` below, this check can go away, since it'll be handled outside the `if/else`.

bobby-brennan:
Looking good! Just one more small change

PR #182, opened 2019-07-31
Wl/webhook test

willfairwinds:


test/webhook_test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+kubectl apply  &> /dev/null -f correctconfig.yaml
+if [ $? -eq 0 ]; then
+    VAR1="pass"
+    echo pass 
+else
+    echo "Test Failed: " 
+    kubectl apply -f correctconfig.yaml
```
bobby-brennan: I think you want to get rid of this line, and add some info more to the `echo` above.

test/webhook_test.sh
```diff
@@ -0,0 +1,24 @@
...
+    VAR2="pass"
+    echo pass 
+else
+    echo "Test Failed: Polaris should have prevented this deployment due to configuration problems."
+fi
+
+if [ $VAR1 = "pass" -a $VAR2 = "pass" ]; then
+    echo "Tests Passed."
+else
+    echo "Tests Failed"
```
bobby-brennan: Let's add `exit 1` below this line, so the script exits with an error.

test/webhook_test.sh
```diff
@@ -0,0 +1,24 @@
...
+
+kubectl apply &> /dev/null -f incorrectconfig.yaml
+if [ $? -ne 0 ]; then
+    VAR2="pass"
+    echo pass 
+else
+    echo "Test Failed: Polaris should have prevented this deployment due to configuration problems."
+fi
+
+if [ $VAR1 = "pass" -a $VAR2 = "pass" ]; then
```
bobby-brennan: These `=` should be `==`

test/webhook_test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+kubectl apply -f https://github.com/FairwindsOps/polaris/releases/latest/download/webhook.yaml
```
bobby-brennan: This should be `./deploy/webhook.yaml`, so we test against the latest deploy configuration.

Let's also add this line before `kubectl apply`:

`sed -ri "s|'(quay.io/reactiveops/polaris:).+'|'\1${CIRCLE_SHA1}'|" ./deploy/webhook.yaml`

You can see a similar line just got added to the `test_kube_dashboard` job. This changes `deploy/webhook.yaml` to use the code that's being checked in, rather than the last release.

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+sed -ri "s|'(quay.io/reactiveops/polaris:).+'|'\1${CIRCLE_SHA1}'|" ./deploy/webhook.yaml
+kubectl apply &> /dev/null -f ./deploy/webhook.yaml
+sleep 15
+
```
bobby-brennan: Instead of using `VAR1` and `VAR2`, I'd suggest setting a variable `PASSING=1` here (before any of the tests), then setting `PASSING=0` in the else statements if any of the tests fails. That'll make it easier for us to add more tests down the line.

test/correctconfig.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: apps/v1
```
endzyme: nitpick: Naming of this file could potentially reflect what it contains, it's currently just "correct config" but you could consider `passing_test.deployment.yaml` to be more expressive and explicit.

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+sed -ri "s|'(quay.io/reactiveops/polaris:).+'|'\1${CIRCLE_SHA1}'|" ./deploy/webhook.yaml
```
endzyme: I'm usually guilty of not doing this enough, but we should always try to comment regex for the intended results (for future reference of intent).

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
...
+kubectl apply &> /dev/null -f ./deploy/webhook.yaml
+sleep 15
```
endzyme: If we can avoid it, I always try to seek a more deterministic approach to asynchronous tasks. In this scenario we're introducing a race condition that's not an actual test. If we aren't ready after 15 seconds, is this an important bit of knowledge for test cases? 

Another approach may be to sleep 15, and write a "check condition" to see if we're in the desired state after that 15 seconds. Another approach (if the "read in 15 seconds" is arbitrary and not a real test) is to just write that test and loop until it passes or timeout after X second so we're not forever stuck checking that the webhook initialized correctly.

Since you have a "known bad config", we can probably use that in a while loop to proceed when the webhook successfully blocks the ability to apply the known bad config.

Happy to discuss more of what I mean here :)

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
endzyme: It's not always 100% necessary but I see lots of people add `set -e` and `set -o pipefail` so the script will exit unclean with any unexpected error and also will catch `piped` errors (in a scenario where a piped shell command can mask an exit code. [More info here](https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail)

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+kubectl apply &> /dev/null -f test/incorrectconfig.yaml
+if [ $? -ne 0 ]; then
+    VAR2="pass"
+    echo pass 
+else
+    echo "Test Failed: Polaris should have prevented this deployment due to configuration problems."
+fi
+
+if [ "$VAR1" == "pass" -a "$VAR2" == "pass" ]; then
```
endzyme: Nit: As future people augment this to add more test cases, adding more of these `and ...` type blocks may become tedious. An alternative is to define a basic framework for your script by defining a variable at the top like `ALL_TESTS_PASSED=true` and then as you run the tests you can call a function if it fails to set `ALL_TESTS_PASSED=false`. Then at the end of the script you can do a simple check:

```shell
if $ALL_TESTS_PASSED; then
  echo "yey!"
else
  echo "booo tests failed"
  exit 1
fi
```

test/webhook_test.sh
bobby-brennan: I don't think we want to `set -e` here, since we're expecting some commands to fail. `set -o pipefail` seems safe though.

You could also `set -e` for the setup, then `set +e` before you start running the tests.

test/webhook_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+sed -ri "s|'(quay.io/reactiveops/polaris:).+'|'\1${CIRCLE_SHA1}'|" ./deploy/webhook.yaml
+kubectl apply &> /dev/null -f ./deploy/webhook.yaml
+sleep 15
+
+kubectl apply &> /dev/null -f test/correctconfig.yaml
```
endzyme: Nit: Although it technically doesn't make a difference, I almost never see this structure in the wild. Normally redirecting output is put at the end of the command for readability. 

`kubectl apply -f test/correctconfig.yaml &>/dev/null` is normally how you would see it in the wild.

test/webhook_test.sh
endzyme: The `set -e` is actually a safer mechanism for testing to catch all unexpected errors. If will converted his tests to be `if kubectl...` instead of `kubectl ... ; if [ $? -ne 0 ]; then...` - it would actually work fine with `set -e`

From the docs: 
>The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !.
[docs](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)

test/webhook_test.sh
bobby-brennan: Nick makes a good point. Here's a safer way to wait for the webhook:
```bash
timeout=15
while ! kubectl get pods -n polaris | grep "polaris-webhook.*Running"; do
  echo "Waiting for webhook to start..."
  if [ $timeout -eq 0 ]; then
    echo "Timed out while waiting for webhook to start"
    exit 1
  fi
  timeout=$((timeout-1))
  sleep 1
done
echo "Webhook started!"
```

test/webhook_test.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+if [ kubectl apply -f test/failing_test.deployment.yaml &> /dev/null  -ne 0 ]; then
+    echo pass 
+else
+    ALL_TESTS_PASSED=false
+    echo "Test Failed: Polaris should have prevented this deployment due to configuration problems."
+fi
+
+if [ $ALL_TESTS_PASSED ]; then
```
bobby-brennan: Sadly bash isn't great with booleans. The tests [are actually failing](https://circleci.com/gh/FairwindsOps/polaris/1780)

You should change `true` to `1` and `false` to `0`, and check `if [ $ALL_TESTS_PASSED -eq 1]`

test/webhook_test.sh
```diff
@@ -0,0 +1,41 @@
...
+ALL_TESTS_PASSED=true
+
+if [ kubectl apply -f test/passing_test.deployment.yaml &> /dev/null eq 0 ]; then
+    echo pass 
+else
+    ALL_TESTS_PASSED=false
+    echo "Test Failed: Polaris prevented a deployment with no configuration issues." 
+fi
+
+if [ kubectl apply -f test/failing_test.deployment.yaml &> /dev/null  -ne 0 ]; then
```
bobby-brennan: I don't think this will quite work. What you want is:

`if ! kubectl apply -f test/failing_test.deployment.yaml &> /dev/null; then`

Same thing above, without the `!`.

Without the `[]`, bash runs this as a command, and it evaluates to true if and only if the command exits with `0`.

deploy/webhook.yaml
```diff
@@ -235,10 +235,12 @@ spec:
         - name: webhook
```
bobby-brennan: Can you revert the changes to this file? `git checkout origin/master deploy/webhook.yaml`

pkg/webhook/validator.go
```diff
@@ -80,6 +80,8 @@ func NewWebhook(name string, mgr manager.Manager, validator Validator, apiType r
 	if err != nil {
 		logrus.Errorf("Error building webhook: %v", err)
 		os.Exit(1)
+	} else {
+		logrus.Info("Webhook started!")
```
bobby-brennan: Can you change this line to `logrus.Info(name + " webhook started")`? We hit this function for several different controller types, so currently it looks like the webhook starts half a dozen times.

test/webhook_test.sh
```diff
@@ -0,0 +1,43 @@
...
+
+#Webhook started, setting all tests as passed initially.
+ALL_TESTS_PASSED=1
+
+if ! kubectl apply -f test/passing_test.deployment.yaml &> /dev/null; then
+    ALL_TESTS_PASSED=0
+    echo "Test Failed: Polaris prevented a deployment with no configuration issues." 
+fi
+
+if kubectl apply -f test/failing_test.deployment.yaml; then
```
bobby-brennan: Probably want to add `&> /dev/null` here too - otherwise the logs get cluttered

test/webhook_test.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo "Waiting for webhook to be ready"
+    while ! kubectl get pods -n polaris | grep -E "webhook.*1/1.*Running"; do
+        check_timeout "${timeout_epoch}"
+        echo -n "."
+    done
+
+    echo "Webhook started!"
+}
+
+function check_dashboard_is_ready() {
```
bobby-brennan: I think you can get rid of this now

test/webhook_test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+function grab_logs() {
+    kubectl -n polaris get pods -oyaml -l app=polaris
+    kubectl -n polaris describe pods -l app=polaris
+    kubectl -n polaris logs -l app=polaris
+}
+
+# Install the webhook and dashboard
+kubectl apply -f ./deploy/webhook.yaml &> /dev/null
+kubectl apply -f ./deploy/dashboard.yaml &> /dev/null
```
bobby-brennan: You can get rid of this line

test/webhook_test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# wait for the webhook and dashboard to come online
+check_webhook_is_ready
+check_dashboard_is_ready
```
bobby-brennan: Ditto

.circleci/config.yml
```diff
@@ -130,7 +130,8 @@ jobs:
     steps:
       - checkout
       - *install_k8s
-      - *test_kube_dashboard
```
bobby-brennan: You can get rid of `test_kube_dashboard` above now

test/kube_dashboard_test.sh
```diff
@@ -0,0 +1,36 @@
+sed -ri "s|'(quay.io/reactiveops/polaris:).+'|'\1${CIRCLE_SHA1}'|" ./deploy/dashboard.yaml
+
+
+function check_dashboard_is_ready() {
+    local timeout_epoch
+    timeout_epoch=$(date -d "+2 minutes" +%s)
+    echo "Waiting for dashboard to be ready"
+    while ! kubectl get pods -n polaris | grep -E "dashboard.*1/1.*Running"; do
+        echo -n "."
```
bobby-brennan: I think you need a `check_timeout "${timeout_epoch}"` above this

PR #188, opened 2019-08-08
Wl/stored audits

willfairwinds:


pkg/dashboard/dashboard.go
```diff
@@ -151,13 +151,16 @@ func GetRouter(c conf.Configuration, auditPath string, port int, basePath string
 		w.Write(favicon)
 	})
 	router.HandleFunc("/results.json", func(w http.ResponseWriter, r *http.Request) {
-		k, err := kube.CreateResourceProvider(auditPath)
-		if err != nil {
-			logrus.Errorf("Error fetching Kubernetes resources %v", err)
-			http.Error(w, "Error fetching Kubernetes resources", http.StatusInternalServerError)
-			return
+		if auditData == nil {
```
bobby-brennan: We still need to handle the case where `auditData != nil`

I think the best thing to do is move `RunAudit` out of `JSONHandler` up to here, then pass the audit to `JSONHandler`.

Then you can move the call to `JSONHandler` outside of this `if` block - so
```go
if auditData == nil {
  k, err := kube.CreateResourceProvider(auditPath)
  // handle err
  auditData, err = validator.RunAudit(c, k)
  // handle err
}
JSONHandler(w, r, c, auditData)
```

main.go
```diff
@@ -27,7 +27,7 @@ import (
 	conf "github.com/fairwindsops/polaris/pkg/config"
 	"github.com/fairwindsops/polaris/pkg/dashboard"
 	"github.com/fairwindsops/polaris/pkg/kube"
-	"github.com/fairwindsops/polaris/pkg/validator"
+	validator "github.com/fairwindsops/polaris/pkg/validator"
```
bobby-brennan: This change shouldn't be necessary - does it work without it?

main.go
```diff
@@ -118,8 +136,8 @@ func main() {
 	}
 }
 
-func startDashboardServer(c conf.Configuration, auditPath string, port int, basePath string) {
-	router := dashboard.GetRouter(c, auditPath, port, basePath)
+func startDashboardServer(c conf.Configuration, auditPath string, port int, basePath string, auditData *validator.AuditData) {
+	router := dashboard.GetRouter(c, auditPath, port, basePath, nil)
```
bobby-brennan: I think this `nil` should be `auditData`

main.go
```diff
@@ -103,7 +104,24 @@ func main() {
 	if *webhook {
 		startWebhookServer(c, *disableWebhookConfigInstaller, *webhookPort)
 	} else if *dashboard {
-		startDashboardServer(c, *auditPath, *dashboardPort, *dashboardBasePath)
+		if *loadAuditFile == "" {
+			startDashboardServer(c, *auditPath, *dashboardPort, *dashboardBasePath, nil)
+		} else {
+			auditData := validator.AuditData{}
```
bobby-brennan: Let's add YAML support using `YAMLOrJSONDecoder`, [as seen here](https://github.com/FairwindsOps/polaris/blob/master/pkg/config/config.go#L141)

pkg/dashboard/dashboard.go
```diff
@@ -151,13 +151,23 @@ func GetRouter(c conf.Configuration, auditPath string, port int, basePath string
...
+			k, err := kube.CreateResourceProvider(auditPath)
+			if err != nil {
+				logrus.Errorf("Error fetching Kubernetes resources %v", err)
+				http.Error(w, "Error fetching Kubernetes resources", http.StatusInternalServerError)
+				return
+			}
+
+			*auditData, err = validator.RunAudit(c, k)
```
bobby-brennan: I think this needs to be something like
```go
auditDataObj, err = validator.RunAudit(c, k)
// handle err
auditData = &auditDataObj
```

main.go
```diff
@@ -103,7 +104,37 @@ func main() {
...
+			oldFileBytes, err := ioutil.ReadFile(*loadAuditFile)
+			if err != nil {
+				logrus.Errorf("Unable to read contents of loaded file: %v", err)
+				os.Exit(1)
+			}
+
+			if err != nil {
```
bobby-brennan: You can delete this error block, since it's already handled above

main.go
```diff
@@ -285,3 +305,17 @@ func runAndReportAudit(c conf.Configuration, auditPath string, outputFile string
 	}
 	return auditData
 }
+func decode(oldFileBytes []byte) (validator.AuditData, error) {
```
bobby-brennan: To keep things clean, can you move this function into `pkg/validator/types.go`, and rename it to `ParseAudit`?

Even better, maybe create a `GetAuditFromFile` function in `pkg/validator/types.go` that contains all the logic above in `main()`.

In general, it's good to keep `main.go` small, since it's not as easy to test/reuse as something in `pkg/`. We've let it get a bit out of hand 😬

main.go
```diff
@@ -31,6 +32,7 @@ import (
 	fwebhook "github.com/fairwindsops/polaris/pkg/webhook"
 	"github.com/sirupsen/logrus"
 	apitypes "k8s.io/apimachinery/pkg/types"
+	yaml2 "k8s.io/apimachinery/pkg/util/yaml"
```
endzyme: Nitpick: Can this be named more appropriately? It's not `yaml` version 2 or something, it's specifically `apiMachineryYAML`. I think it would be beneficial to future readers to understand why there is a `yaml` and a `yaml2`. Non-blocking comment.

main.go
```diff
@@ -59,6 +61,7 @@ func main() {
 	auditOutputURL := flag.String("output-url", "", "Destination URL to send audit results")
 	auditOutputFile := flag.String("output-file", "", "Destination file for audit results")
 	auditOutputFormat := flag.String("output-format", "json", "Output format for results - json, yaml, or score")
+	loadAuditFile := flag.String("load-audit-file", "", "Runs the dashboard with data saved from a past audit.")
```
endzyme: Maybe not a good comment for this PR: but noteworthy. The arguments for this one binary are getting very complex and intertwined. It's technically possible to set `-load-audit-file` with the `-webhook` flag and it will just ignore the first flag. This (to me) is confusing and maybe could leverage some subcommand support in the future. Again not for this PR but something to consider a future change for.

main.go
endzyme: Agreed, the main.go is getting a bit long and has lots of logic in it. I also agree the name the function could use some love. `ParseAudit` is ok with me, but I'm also very verbose when it comes to function names and `ReadAuditFromFile()` usually is something I would be explicit about.

main.go
```diff
@@ -103,7 +106,24 @@ func main() {
 	if *webhook {
 		startWebhookServer(c, *disableWebhookConfigInstaller, *webhookPort)
 	} else if *dashboard {
-		startDashboardServer(c, *auditPath, *dashboardPort, *dashboardBasePath)
+		if *loadAuditFile == "" {
```
endzyme: I'm down to have a quick working session if you want to go back and forth on some ideas of where this logic should live. I am currently thinking that this should all be in the dashboard pkg and you should always pass `loadAuditFile` to `startDashboardServer()`. 

I'm happy to pair up and talk through your reasoning for this so I can better understand the problem.

main.go
bobby-brennan: 100% agree. https://github.com/FairwindsOps/polaris/issues/190

main.go
bobby-brennan: Even though this functionality is currently only used by the dashboard, loading an audit from a file doesn't really seem like a dashboardy thing to do.

I'd suggest putting the loadAuditDataFromFile fn in the same package where the AuditData type is defined, i.e. `pkg/validator`

endzyme: I think we're saying the same thing but I haven't been super clear. I mean the actual marshalling of the file logic should definitely be in the validator pkg but calling the `LoadAuditFromFile()` should happen in dashboard instead of main. If we're not saying the same thing I'm happy to connect in slack :)

bobby-brennan: Ah no, that SGTM

bobby-brennan:
Looks good! Just remove the file at `pkg/test.json`

pkg/validator/types.go
```diff
@@ -197,3 +204,34 @@ type ResultMessage struct {
...
+	auditData := AuditData{}
+	oldFileBytes, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		logrus.Errorf("Unable to read contents of loaded file: %v", err)
+		os.Exit(1)
+	}
+	auditData, err = ParseAudit(oldFileBytes)
+	if err != nil {
+		logrus.Errorf("Error parsing file contents into auditData: %v", err)
+		os.Exit(1)
```
endzyme: I think it would be more graceful if this ReadAuditFromFile was able to return an error that could be caught in main.go. Since this is a library file (effectively not the command runner) then it's my opinion that the main.go should be the ultimate decider of `os.Exit()` type functionality.

pkg/validator/types.go
```diff
@@ -197,3 +204,34 @@ type ResultMessage struct {
 	Type     MessageType
 	Category string
 }
+
+// ReadAuditFromFile reads the data from a past audit stored in a JSON or YAML file.
+func ReadAuditFromFile(fileName string) AuditData {
+	auditData := AuditData{}
+	oldFileBytes, err := ioutil.ReadFile(fileName)
```
endzyme: Is there a need here to use ioutil.ReadFile to first read the bytes into memory before running `ParseAudit`? Is this split up for testing purposes or could this be made more concise by using `os.Open(fileName)` and passing that to the `NewYAMLOrJSONDecoder()` function? `os.Open()` Returns a `File` type which conforms to the `io.Reader` interface.

endzyme:
Oops looks like this was already merged while i was looking at it - no worries :+1: :shipit:

PR #189, opened 2019-08-12
sleep to ensure dashboard/webhook are ready

bobby-brennan:


endzyme:
Is this something that's just impossible to determine in code or is this a band-aid for a longer solution that we have an open issue for?

bobby-brennan:
For the webhook, there might be a better solution if we add a readiness listener.

For `go run main.go --dashboard`, we might be able to add some log output, and wait for that to appear.

For the dashboard, where we have to run `port-forward`, I'm not sure if there's a good solution, but we might be able to wait for log output there as well.

bobby-brennan:
Opened https://github.com/FairwindsOps/polaris/issues/191 to track

PR #194, opened 2019-08-29
add images to test yaml and json config

kimschles:
This PR adds the image config to our test data to ensure that the YAML and JSON parsing still work properly. All tests pass with the addition.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -58,6 +58,15 @@ resources:
     warning:
       below: 300M
       above: 4G
+images:
+  tagNotSpecified: error
+  pullPolicyNotAlways: warning
+  whitelist:
+    error:
+      - gcr.io/*
+  blacklist:
+    warning:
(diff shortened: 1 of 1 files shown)


pkg/config/config_test.go
```diff
@@ -58,6 +58,15 @@ resources:
     warning:
       below: 300M
       above: 4G
+images:
+  tagNotSpecified: error
+  pullPolicyNotAlways: warning
+  whitelist:
+    error:
+      - gcr.io/*
```
endzyme: Thanks for adding these @kimschles! It looks as though you're setting them in the config, and the config parses correctly, but we're not actually validating the settings are parsed correctly in `testParsedConfig()`. This means, the YAML/JSON will be read, but it's not actually checking the values against anything serialized into Go structs.

I am going to write up an example here (below) on how one might achieve that, to assure we're loading what we expect to load from the yaml & json.

PR #197, opened 2019-09-09
add IDs to each check

rbren:


pkg/validator/ids.go
```diff
@@ -0,0 +1,18 @@
+package validator
+
+import (
+	"reflect"
+)
+
+func getIDFromField(config interface{}, name string) string {
+	t := reflect.TypeOf(config)
```
endzyme: Have we performance tested this reflection function on lookups? I have heard that adding this type of type testing is a high cost to performance, but not sure on what scale. If we had 1,000 checks would it be such a marginal difference that it wouldn't be notable? (Also there's the weight of testing the performance itself, cost of testing this might outweigh the value of knowing if it is a problem) :)

rbren: This function takes on average ~1.634e-06 seconds to run, so I think we're OK :)

endzyme: way too long - :) - thanks for looking into it and teaching me something. Out of curiosity, do you have a snippet of how you tested this out?

rbren: Here's what I did:

```go
start := time.Now()
// do some stuff
end := time.Now()
fmt.Println("Took %f seconds", end.Sub(start).Seconds())
```

PR #200, opened 2019-09-11
Add source installation/go get instructions to usage documentation.

ayang64:
Polaris can easily be installed using "go get".  We should make that a prominent way of installing it.


ayang64:
Thank you -- absolutely.  I'll revise this patch ASAP.

endzyme:
Looks like #202 is in - reviewing this PR now

rbren:
Talked about this internally - there doesn't seem to be a way to pin the version to a stable release, so we don't want to make `go get` a standard install path.

Thanks for the idea though!

Issue #132, opened 2019-06-04
static content is not loaded when using nginx-ingress with custom base path


troian:
We would like to get polaris dashboard accessible from subpath of our domain (e.g. `https://example.com/polaris`)

Issue is static content is not loaded as deployment expects them without `polaris` path.

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: example.com-polaris
  namespace: polaris
  annotations:
    kubernetes.io/ingress.provider: "nginx"
    kubernetes.io/ingress.class: "nginx-ks"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/secure-backends: "false"
    nginx.ingress.kubernetes.io/rewrite-target: "/$1"
    nginx.ingress.kubernetes.io/configuration-snippet: |
      rewrite ^(/polaris)$ $1/ permanent;
spec:
  tls:
    - hosts:
        - example.com
      secretName: star-example-com
  rules:
    - host: cluster.notifai.io
      http:
        paths:
          - path: /polaris/?(.*)
            backend:
              serviceName: polaris-dashboard
              servicePort: 8080
```

bobby-brennan:
Thanks for the report! This was fixed in `0.1.5`.

To serve the dashboard on a base path like `/polaris`, you'll need to set `--dashboard-base-path="/polaris/"`

#136 will add this as an option to the helm chart. You can also alter the file at `./deploy/dashboard.yaml`.

Hope that helps! Let me know if you have any trouble.

troian:
Does not seem to work unfortunately.
When trying to access dashboard [within internal link](http://polaris-dashboard.polaris.svc.cluster.local) without custom path it works fine

adding `--dashboard-base-path="/polaris/"` breaks that so neither `example.com/polaris/` nor [internal link](http://polaris-dashboard.polaris.svc.cluster.local) works

rbren:
@troian - sorry, I missed your follow-up message. I've reopened this issue for now.

When I set up an Ingress similar to yours using nginx-ingress (config below), combined with the `--dashboard-base-path` flag, example.com/polaris works as expected (though accessing the service via `kubectl port-forward` doesn't work).

We have an [open PR](https://github.com/FairwindsOps/polaris/pull/201#issuecomment-535693440) that would mount the router on the path provided in `--dashboard-base-path`, which I think would obviate the need for the rewrite rule, and allow both the service and the external URL to work.

[...]

PR #201, opened 2019-09-11
pkg/dashboard: setup basePath as a path prefix in routing

adamdecaf:
Awesome project! It works for me with a `port-forward`, but not with `-dashboard-path-prefix`. After this change I can load the dashboard with a base path (`/polaris/` in my case), but I haven't tested building a Docker image for my cluster yet. 

~~Btw, I was unable to sign the CLA at https://cla-assistant.io/fairwinds/polaris (404)~~ With `?pullRequest=201` I could sign the CLA.

### DIFF
--- a/pkg/dashboard/dashboard.go
+++ b/pkg/dashboard/dashboard.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"html/template"
 	"net/http"
+	"path"
 	"strings"
 
 	conf "github.com/fairwindsops/polaris/pkg/config"
@@ -137,7 +138,7 @@ func writeTemplate(tmpl *template.Template, data *templateData, w http.ResponseW
 
 // GetRouter returns a mux router serving all routes necessary for the dashboard
 func GetRouter(c conf.Configuration, auditPath string, port int, basePath string, auditData *validator.AuditData) *mux.Router {
-	router := mux.NewRouter()
+	router := mux.NewRouter().PathPrefix(basePath).Subrouter()
 	router.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("OK"))
 	})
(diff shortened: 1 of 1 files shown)


rbren:
Thanks for the details Adam! Sorry for the delay in response.

I think to support your use case in a backward-compatible way, we'd have to create a new flag, maybe `dashboard-mount-path`, which would get passed in alongside `basePath`. If you want to modify your PR to do that we can merge it in, or we can take a look at adding support for this in our next release.

adamdecaf:
What about `-dashboard-web-path[-prefix]`?

rbren:
Sounds good, I like `-dashboard-web-path-prefix`

adamdecaf:
Reading this again it seems this PR fixes the problem ran into at https://github.com/FairwindsOps/polaris/issues/132, so I'm not sure why we need a new flag. `--dashboard-path-prefix` is already passed to `dashboard.GetRouter` so we want to merge two "path prefix" flags together?

rbren:
🤔 you may be right. I never saw the follow-up message from troian there.

My main concern was breaking his use-case if we merged this. Seeing as it was never fixed, I'm less concerned. Will give him a day or so to respond, otherwise I'll assume this is good to go.

adamdecaf:
cc @troian thoughts?

troian:
Just in case I miss conversation point. Just check with version `0.4.0`. Config is still the same as in #132

adamdecaf:
With 0.4.0 and your config from #132 the static assets are still not rendered properly? I was able to specify `--dashboard-path-prefix` and load under a `/polaris/` path locally.

troian:
Yes. I checked 0.5.0-beta2 and that gave me same result.
Sounds like something wrong then on our side

rbren:
@troian can you try using `0.5.0-beta3`? That incorporates this change. I think you'll also need to remove the rewrite rule in your ingress.

adamdecaf:
I adjusted the patch a bit and it works better locally for me now. The only annoying thing is `/health` doesn't respond with a path. Instead you'd need `/polaris/health`.

rbren:
Great! I don't think there's a good way around needing to adjust the health check, mounting `/health` on the base path and everything else on `/polaris` doesn't seem quite right. We'll adjust the liveness/readiness checks in the helm chart accordingly.

Thanks for making this change Adam, will merge now

PR #202, opened 2019-09-12
skip kubernetes tests for forked PRs

rbren:
Forked PRs don't build the docker images and push to quay, so we can't run the k8s tests.

This also moves some test files to a separate directory


rbren:
Yeah this is a tough call. To be clear, we'll still run `go test` for forked PRs, we just miss this e2e test. The e2e tests will also run after merging, and our badge will turn red if they broke.

The reason is that we can't push images to quay for forked PRs without exposing credentials, and the e2e tests need to run against the image associated with this commit.

I often find it frustrating in dev/testing that I can't just push a local image to k8s, and instead need to first push it to a middleman registry. I'd love it if setting `image: localhost/polaris:latest` just magically pushed my image to the cluster.

The "solution" is probably something like
* add an image registry to the KIND cluster
* build the image on CircleCI
* push the image to the cluster
* point the deploy configs to the cluster-hosted image

PR #203, opened 2019-09-13
fix empty dashboard when load-audit-file not specified

rbren:


rbren:
Thanks for checking Kim! Ptr is short for pointer - more info here: https://golangbot.com/pointers/

PR #204, opened 2019-09-13
Add exemptions to config

rbren:


rbren:
Screen Shot 2019-10-15 at 10 01 08 AM

endzyme:
Initial testing

Tested:
- Ignore all errors for all scanned deployments (results in 100% score) (seems fine)
- Using with `examples/config-full.yaml` and got very weird results (see screenshot) (bug?)
- Tried to figure out how to exclude capabilities individually (can't right now) but wasn't immediately clear how to look up the rule names
- Tested invalid configuration for exemptions (tried a non-valid rule name but didn't fail as I might expect) - accidentally had a typo in a rule name and it just skips over it


Initial thoughts: Looks like maybe the reflection should have the context of the "group" since it is using the `` `json:"<name>"` `` as the name of the rule. I think you might accidentally exempt something if there happened to be a duplicate name in the json formatting options. 

Example:
```golang
type Configuration struct {
...
	Security           Security              `json:"security"`
	AnotherThing       TheThing              `json:"theThing"`
	Exemptions         []Exemption           `json:"exemptions"`
...
}

type Security struct {
...
	Capabilities SecurityCapabilities `json:"capabilities"`
...
}

type AnotherThing struct {
...
	Capabilities []string `json:"capabilities"`
...
}
---------
```
Coupled with:
```yaml
exemptions:
- controllerNames:
  - some-controller-name-here
  rules:
  - capabilities
```
This would accidentally exclude both the "security" context group as well as the AnotherThing context group. (if i'm reading the code right).
```

rbren:
Good points! Thanks for the review. Can definitely try and hit some of these (esp failing on invalid names) in a future PR.

PR #205, opened 2019-09-18
update to 0.5.0

rbren:


rbren:
Not sure why `nodeSelector` and `tolerations` changed (maybe I updated helm?), but `kubectl` doesn't seem to mind.

deploy/dashboard.yaml
```diff
@@ -216,8 +216,8 @@ spec:
           subPath: config.yaml
           readOnly: true
       serviceAccountName: polaris-dashboard
-      nodeSelector: {}
-      tolerations: []
+      nodeSelector:
+      tolerations:
```
endzyme: are these changes intentional - it's not mentioned in the commit message and I wanted to assure it wasn't an IDE formatting thing so I can make sure mine is setup the same way

rbren: I think this is due to a Helm update (I'm on `2.13.0` now)

Issue #37, opened 2019-03-29
Add Validation for Image Pattern Matching


robscott:
We should have a list of patterns that an image attribute must match one of, or a list of patterns that an image attribute should not match any of. Generally these would never be set at the same time, but I can't think of any reason we need to add any logic to ensure that they're never both set. I think the most sensible approach here is to ensure that `container.image` matches at least one pattern defined in the whitelist or does not match any patterns defined in the blacklist. We do not want to do full regex matching here, but supporting a `*` wildcard character would simplify things. Ideally we could create a small method that determined if a blacklist or whitelist item matched a string, and then write a bunch of unit tests to cover use cases like:

`quay.io/reactiveops*` 
- `quay.io/reactiveops` (invalid)
- `quay.io/reactiveops/` (valid)
- `quay.io/unreactiveops` (invalid)
- `quay.io/reactiveops/rbac-manager` (valid)

`quay.io/*ops*` 
- `quay.io/reactiveops` (invalid)
- `quay.io/reactiveops/` (valid)
- `quay.io/unreactiveops` (valid)
- `quay.io/reactiveops/rbac-manager` (valid)

`quay.io/reactiveops/rbac-manager` 
- `quay.io/reactiveops` (invalid)
- `quay.io/reactiveops/` (invalid)
- `quay.io/unreactiveops` (invalid)
- `quay.io/reactiveops/rbac-manager` (valid)

towolf:
> We do not want to do full regex matching here

Just a note, if you will. We had this discussion ourselves 2 years ago and we agreed that anybody who is interested in policing images is likely to be well versed in regex. So we just went with it:

https://github.com/freach/docker-image-policy-plugin/#plugin-configuration

Sometimes the regex is really necessary:

```
image-policy:                                            
  defaultAllow: false                                    
                                                         
  blacklist: []                                          
    # - ':latest$'                                       
    # - ':master$'                                       
                                                         
  whitelist:                                             
    # allow anything from gitlab registry                
    - '^gitlab\.example\.com:5005'                         
                                                         
    # not sure we need these at all                      
    # only for one-off usage                             
    - '^alpine(:|$)'                                     
    - '^busybox(:|$)'                                    
    - '^maven(:|$)'     # used by platform for building  
    - '^memcached(:|$)' # used by socorro                
    - '^mysql(:|$)'     # used by gitlab ci/cd "service" 
    - '^nginx(:|$)'                                      
    - '^node(:|$)'                                       
[...]

PR #208, opened 2019-09-30
Add image registry check

kimschles:
**What does this PR do?**
This PR gives polaris the ability to check a container's image against a whitelist and blacklist of image repositories. If the image repo does not match a pattern(s) in the whitelist, polaris should return an error. If the image repo matches a pattern(s) on the blacklist, it should return a warning. 

**Why do we want to do that?** 
This PR addresses [issue #37](https://github.com/FairwindsOps/polaris/issues/37)
We want users to be able to define a list of allowed and blocked image repositories. 

**What high level changes did you make to the code to accomplish that goal?** 
* I added whitelist and blacklist tests to the`testParsedConfig` function in `confit_test.go` file. 
* I added the registry check code to the `validateImage` function in `container.go`. 
* I added `ImageRegistryFailure` and `ImageRegistrySuccess` to the messages package.
* I updated the `docs/check-documentation/images.md` to describe how this functionality works. 

**What other information should the reviewer be aware of when looking at this code?** 
I wasn't able to figure out how to write tests for this functionality, so I'm not certain it works, even though the code in this PR passes CI. I would like to pair up with someone to write the tests. Also, I would like to continue working on this code to address the following: 

To Do: 
* Write tests for this functionality in the `TestValidateImage` function in `container_test.go` 
* Rename whitelist/blacklist to allowed/blocked 
* Refactor the registry check code so that it is less redundant. 

If you'd prefer to review this PR with me via Zoom instead of a written review, I'm cool with that.


pkg/validator/container.go
```diff
@@ -182,6 +184,54 @@ func (cv *ContainerValidation) validateImage(imageConf *conf.Images) {
...
+			}
+		}
+
+	}
+	if hasBlacklistCheck && hasBlacklistFailure == false {
+		cv.addSuccess(messages.ImageRegistrySuccess, category, blacklistCheckID)
+	}
+	for _, imageList := range imageLists {
+		var severity config.Severity
+		if &imageList == &imageConf.Blacklist.Error {
```
rbren: There are a few instances of `blacklist` in this block that need to become `whitelist`

pkg/validator/messages/messages.go
```diff
@@ -96,4 +96,8 @@ const (
 	HostNetworkFailure = "Host network should not be configured"
 	// HostNetworkSuccess message
 	HostNetworkSuccess = "Host network is not configured"
+	// ImageRegistryFailure message
+	ImageRegistryFailure = "The image registry is not allowed or must match the allowed registries list"
```
rbren: Let's break out separate messages for whitelist and blacklist, since they have separate IDs

ejether:
I'm curious about `If the image repo matches a pattern(s) on the blacklist, it should return a warning`. Can you help me understand why a black listed image registry only produces a warning?

kimschles:
> I'm curious about `If the image repo matches a pattern(s) on the blacklist, it should return a warning`. Can you help me understand why a black listed image registry only produces a warning?

That's a good question, and I did my best to understand Rob's intent for the whitelist/blacklist feature by reading the issue he filed, as well as this block of code he wrote: https://github.com/fairwindsops/polaris/blob/master/examples/config-full.yaml#L37-#L42

I'm open to changing it to do what's best for the project.

pkg/validator/messages/messages.go
kimschles: 👍 Updated so that blacklist and whitelist both have failure and success messages.

pkg/validator/container.go
kimschles: Updated.

kimschles:
> > I'm curious about `If the image repo matches a pattern(s) on the blacklist, it should return a warning`. Can you help me understand why a black listed image registry only produces a warning?
> 
> That's a good question, and I did my best to understand Rob's intent for the whitelist/blacklist feature by reading the issue he filed, as well as this block he wrote: https://github.com/fairwindsops/polaris/blob/master/examples/config-full.yaml#L37-#L42. When I wrote tests for that configuration, it seemed like something that doesn't match the whitelist gets an error, while something that matches the blacklist gets a warning.
> 
> I'm open to changing it to do what's best for the project.

I just talked with Robert, and it should be fully configurable.

docs/check-documentation/images.md
```diff
@@ -6,11 +6,14 @@ key | default | description
 ----|---------|------------
 `images.tagNotSpecified` | `error` | Fails when an image tag is either not specified or `latest`.
 `images.pullPolicyNotAlways` | `ignore` | Fails when an image pull policy is not `always`.
+`images.whitelist` | `error` | If the container image does not match the pattern in the whitelist, it will fail with an error.
```
endzyme: Thanks for the work here and docs! I think I am a little confused since it looks like the code supports both warning or error for lists of black/white lists. I think the docs could reflect that by maybe adding "`error` or `warning` (configurable)" to this table. Open to suggestions though!

pkg/config/config_test.go
```diff
@@ -190,4 +220,12 @@ func testParsedConfig(t *testing.T, config *Configuration) {
...
 	controllersToScan := config.ControllersToScan
 	assert.ElementsMatch(t, []SupportedController{Deployments, StatefulSets, Jobs, CronJobs, DaemonSets, ReplicationControllers}, controllersToScan)
+
+	images := config.Images
+	assert.Equal(t, SeverityWarning, images.PullPolicyNotAlways)
+	assert.Equal(t, images.TagNotSpecified, SeverityError)
+	assert.Equal(t, images.Whitelist.Error, []string{"gcr.io/*"})
+	assert.Equal(t, images.Whitelist.Warning, []string(nil))
+	assert.Equal(t, images.Blacklist.Error, []string(nil))
+	assert.Equal(t, images.Blacklist.Warning, []string{"docker.io/*"})
```
endzyme: Thanks so much for adding configuration tests. Are there any other tests coming in a later commit to test the logic of raising warnings or errors when a white/blacklist item is found? I would also be interested in the condition when you set a blacklist and whitelist item to the same thing, which one takes precedence?

pkg/validator/messages/messages.go
```diff
@@ -96,4 +96,12 @@ const (
 	HostNetworkFailure = "Host network should not be configured"
 	// HostNetworkSuccess message
 	HostNetworkSuccess = "Host network is not configured"
+	// RegistryWhitelistFailure message
+	RegistryWhitelistFailure = "The image registry must match the allowed registries list"
```
endzyme: Nit: Can we add the verbiage of `... allowed registry whitelist...` just so the term whitelist is in this failure? (not a huge deal though)

ejether:
> > > I'm curious about `If the image repo matches a pattern(s) on the blacklist, it should return a warning`. Can you help me understand why a black listed image registry only produces a warning?
> > 
> > 
> > That's a good question, and I did my best to understand Rob's intent for the whitelist/blacklist feature by reading the issue he filed, as well as this block he wrote: https://github.com/fairwindsops/polaris/blob/master/examples/config-full.yaml#L37-#L42. When I wrote tests for that configuration, it seemed like something that doesn't match the whitelist gets an error, while something that matches the blacklist gets a warning.
> > I'm open to changing it to do what's best for the project.
> 
> I just talked with Robert, and it should be fully configurable.

Thank makes sense. Thanks!

pkg/validator/messages/messages.go
kimschles: yes. :)

docs/check-documentation/images.md
kimschles: Fixed in the latest commit.

pkg/config/config_test.go
kimschles: Robert helped me add tests for the whitelist/blacklist checks in pkg/validator/container_test.go. 

Regarding your questions about what happens when a blacklist and whitelist item or warnings/errors are set as the same thing, I'm not sure. I'll schedule some time with your or Robert to make a decision on this.

pkg/validator/container.go
```diff
@@ -182,6 +184,57 @@ func (cv *ContainerValidation) validateImage(imageConf *conf.Images) {
...
+
+	// Check whitelist
+	for _, imageList := range imageWhiteLists {
+		var severity config.Severity
+		if imageList == &imageConf.Whitelist.Error {
+			severity = config.SeverityError
+		} else {
+			severity = config.SeverityWarning
+		}
+		for _, imagePattern := range *imageList {
```
rbren: Realizing now we didn't quite get this right. As it stands, the image has to pass _every_ whitelist check, so you couldn't specify e.g. `["docker.io/.*", "quay.io/.*"]`. Instead, we should fail only if _all_ the whitelist checks fail.

I think we just want to change `hasWhitelistFailure` to `hasWhitelistSuccess`. Then we can run
```go
for _, imagePattern := range *imageList {
  hasWhitelistCheck = true
  if regexp.MustCompile(imagePattern).MatchString(cv.Container.Image) {
    hasWhitelistSuccess = true
    break
  }
}
if hasWhitelistCheck {
  if hasWhitelistSuccess{
    addSuccess()
  } else {
    addFailure()
  }
}
```

rbren:
We've implemented this as a custom check now, so I'm going to close this.

PR #213, opened 2019-10-14
don't try to fetch kube resources if loading from file

rbren:


rbren:
Apparently I implemented this in another PR that was already merged. 🤷‍♂

Issue #209, opened 2019-10-03
Don't check for health checks on Jobs/CronJobs


rbren:
### Polaris Version
latest

### Expected Behavior
CronJobs and Jobs should not fail checks for liveness/readiness probes

### Actual Behavior
Lacking liveness/readiness probes on Jobs/CronJobs causes a decrease in the score

ejether:
Help me understand the logic behind Jobs and CronJobs not needing liveness/readiness probes?

davidkarlsen:
@ejether They run a command until completion. The exit code will tell how it went. Readiness probes are used to let pods enter load-balancing - which cronjob's don't do? Same for liveness - I don't see the need for it - the pod simply runs to completion returning some exit-code.

PR #216, opened 2019-10-17
skip health checks for jobs, cronjobs, and initContainers

### LINKED ISSUE #215: Liveness probes should be ignored on initContainers
Adding Livenessprobe to InitContainers results in failures. Are these even supported by k8s?

rbren:
Fixes #215 and #209 

Also skips resource checks if they're not in the config

### DIFF
--- a/pkg/validator/container.go
+++ b/pkg/validator/container.go
@@ -38,7 +38,7 @@ type ContainerValidation struct {
 //       relevant podSpec in order to check certain aspects of a containerSpec.
 //       Perhaps there is a more ideal solution instead of attaching a parent
 //       podSpec to every container Validation struct...
-func ValidateContainer(container *corev1.Container, parentPodResult *PodResult, controllerName string, conf *config.Configuration, isInit bool) ContainerResult {
+func ValidateContainer(container *corev1.Container, parentPodResult *PodResult, conf *config.Configuration, controllerName string, controllerType config.SupportedController, isInit bool) ContainerResult {
 	cv := ContainerValidation{
 		Container:          container,
 		ResourceValidation: &ResourceValidation{},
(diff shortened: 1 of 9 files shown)


pkg/validator/container.go
```diff
@@ -38,7 +38,7 @@ type ContainerValidation struct {
 //       relevant podSpec in order to check certain aspects of a containerSpec.
 //       Perhaps there is a more ideal solution instead of attaching a parent
 //       podSpec to every container Validation struct...
-func ValidateContainer(container *corev1.Container, parentPodResult *PodResult, controllerName string, conf *config.Configuration, isInit bool) ContainerResult {
+func ValidateContainer(container *corev1.Container, parentPodResult *PodResult, conf *config.Configuration, controllerName string, controllerType config.SupportedController, isInit bool) ContainerResult {
```
jordandoig: More of a golang question here, is there a reason the argument order here needed to be changed?

pkg/validator/pod.go
```diff
@@ -27,7 +27,7 @@ type PodValidation struct {
 }
 
 // ValidatePod validates that each pod conforms to the Polaris config, returns a ResourceResult.
-func ValidatePod(conf config.Configuration, controllerName string, pod *corev1.PodSpec) PodResult {
+func ValidatePod(conf config.Configuration, pod *corev1.PodSpec, controllerName string, controllerType config.SupportedController) PodResult {
```
jordandoig: Same as above, re: argument order

pkg/validator/container.go
rbren: Definitely wasn't necessary. I merged in the change that added the `controllerType` parameter, and realized `controllerType` and `controllerName` should probably be next to each other.

pkg/validator/pod.go
rbren: Ditto - just an aesthetic preference

PR #226, opened 2019-12-04
Full support for validating webhook, now that tests are in place

rbren:


makoscafee:
LGTM. there is CLA issue i think

rbren:
CLA has been giving us a lot of trouble lately :/

PR #230, opened 2019-12-06
move to using fairwinds quay repo

rbren:


rbren:
Yup, you were right! Thanks

PR #233, opened 2019-12-19
circleci: set env for go mod

rbren:


.circleci/config.yml
```diff
@@ -132,6 +135,7 @@ jobs:
       - image: circleci/golang:1.12
```
makoscafee: would this make sense if updated to golang1.13 too as in `Dockerfile`

rbren: Good call!

PR #239, opened 2020-01-28
Exclude individual checks from a controller

baderbuddy:
Added the ability to exclude a controller from individual checks via annotation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # x.x.x (next release)
 
+# 0.6.1
+* Added the ability to exempt a particular controller from a particular check.
+
 # 0.6.0
 * Fixed webhook support in Kubernetes 1.16
   * this also removes support for 1.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,4 @@
 # x.x.x (next release)
-
-# 0.6.1
 * Added the ability to exempt a particular controller from a particular check.
 
 # 0.6.0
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # x.x.x (next release)
 
+# 0.6.1
```
rbren: We're actually working on `1.0.0`! You can drop the heading and leave this under the `x.x.x` section so we don't forget it.

We'll do the version bump after some more refactoring (specifically the cmd stuff - e.g. I think we'll replace `polaris --dashboard --dashboard-port=3000` with `polaris dashboard --port=3000`). I'll have to spend some time documenting all the changes and the upgrade path.

docs/usage.md
```diff
@@ -76,6 +76,11 @@ exemptions:
       - hostNetworkSet
 ```
 
+To exempt a controller from a particular check via annotations, use an annotation in the form of `polaris.fairwinds.com/<check>-exempt=true`, e.g.
+```
+kubectl annotate deployment my-deployment polaris.fairwinds.com/cpu-requests-missing-exempt=true
```
rbren: Any reason not to use camelCase like we do for the IDs? From what I can tell `[A-Z]` is legal.

e.g. `cpuRequestsMissing-exempt`

CHANGELOG.md
baderbuddy: Oh, to avoid duplicate work take a look at this, https://github.com/FairwindsOps/polaris/blob/bb/cobra-cmd/docs/usage.md

docs/usage.md
baderbuddy: Well that'll teach me to make assumptions. I don't know why I had it in my head that annotations had to follow the same rules as namespaces or other objects. I'll change it up.

Issue #206, opened 2019-09-18
Fix score when denominator is zero


rbren:
When polaris has nothing to check (e.g. all checks disabled), the score shows up as `9223372036854775808`

PR #240, opened 2020-01-31
Adding support for third-party Controllers (e.g. OpenShift)

baderbuddy:
When pulling ReplicationController it will now look for an OwnerReference and roll up duplicates to that owner. This allows Polaris to work with OpenShift DeploymentConfigs but also any other CRD.

Also added a quick check for dividing by 0.


pkg/validator/summary.go
```diff
@@ -17,6 +17,9 @@ type CountSummaryByCategory map[string]CountSummary
 // GetScore returns an overall score in [0, 100] for the CountSummary
 func (cs CountSummary) GetScore() uint {
 	total := (cs.Successes * 2) + cs.Warnings + (cs.Errors * 2)
+	if total == 0 {
```
rbren: Thanks! I can't believe [this bug is still open](https://github.com/FairwindsOps/polaris/issues/206) 😬

pkg/validator/controller.go
```diff
@@ -32,24 +33,60 @@ func ValidateController(conf *conf.Configuration, controller controller.Interfac
...
+	return results
+}
+
 // ValidateControllers validates that each deployment conforms to the Polaris config,
 // builds a list of ResourceResults organized by namespace.
 func ValidateControllers(config *conf.Configuration, kubeResources *kube.ResourceProvider) ([]ControllerResult, error) {
 	var controllersToAudit []controller.Interface
 	for _, supportedControllers := range config.ControllersToScan {
-		loadedControllers, _ := controllers.LoadControllersByKind(supportedControllers, kubeResources)
+		loadedControllers, err := controllers.LoadControllersByKind(supportedControllers, kubeResources)
```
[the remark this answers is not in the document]
baderbuddy: It looks like Jobs and CronJobs were throwing some kind of error, I'd hate to return the error and kill everything based on what could be a mistake in the config. So I think logging makes sense.

pkg/validator/controller.go
```diff
@@ -32,24 +33,60 @@ func ValidateController(conf *conf.Configuration, controller controller.Interfac
...
+		Namespace:   controller.GetObjectMeta().Namespace,
+		Results:     ResultSet{},
+		PodResult:   podResult,
+		CreatedTime: controller.GetObjectMeta().CreationTimestamp.Time,
+	}
+	owners := controller.GetObjectMeta().OwnerReferences
+	// If an owner exists then set the name to the controller.
+	// This allows us to handle CRDs creating Controllers or DeploymentConfigs in OpenShift.
+	if len(owners) > 0 {
+		firstOwner := owners[0]
```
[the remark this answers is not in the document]
baderbuddy: None, I hope it's not possible but the types seem to technically allow it.

baderbuddy:
What was the original thinking with the ControllersToScan option? That might help me in thinking through how to fit this in. I was trying to think through any kind of scenario where I wouldn't just want to scan every kind of controller that I could.

rbren:
TBH, I don't fully remember 😅 - the guy who implemented it is no longer on the project. I wouldn't be opposed to removing it for v1.

I still have some concerns over the current implementation though - e.g. I think we'll end up seeing multiple `CronJob` objects - one for the original, and one for each `Job` it's created. And if the ownership is XController -> Pod, with no intermediate controller type, we'll miss it.

I do dislike our current pattern of needing a bunch of boilerplate code for each controller type, but I think it'd be good to stay consistent between vanilla k8s and open shift. Maybe we can sync offline and come up with a better path forward. I think checking pod parents like you're doing here might be an interesting route.

baderbuddy:
Robert, do you want to take another look? I think I have the dynamic querying for parents working, I even have some dynamicness working for loading from a YAML file. I think this is at least the right track, but there's probably more to clean up.

pkg/validator/controllers/interface.go
```diff
@@ -13,16 +11,33 @@ import (
 type Interface interface {
```
rbren: Do we still need an interface, if `GenericController` is the only implementer?

pkg/kube/resources.go
```diff
@@ -54,18 +50,12 @@ func CreateResourceProvider(directory string) (*ResourceProvider, error) {
...
-		CronJobs:               []batchv1beta1.CronJob{},
-		ReplicationControllers: []corev1.ReplicationController{},
-		Namespaces:             []corev1.Namespace{},
-		Pods:                   []corev1.Pod{},
+		ServerVersion: "unknown",
+		SourceType:    "Path",
+		SourceName:    directory,
+		Nodes:         []corev1.Node{},
+		Namespaces:    []corev1.Namespace{},
+		Pods:          []corev1.Pod{},
```
rbren: Rather than `Pods`, maybe it makes sense for this to be `Controllers: []GenericController{}`. Then we can do all the Owner resolution here, instead of inside `Validate`.

If possible, I'd like to keep interactions w/ the k8s API here, rather than passing around an interface.

pkg/kube/resources.go
```diff
@@ -168,222 +137,77 @@ func CreateResourceProviderFromAPI(kube kubernetes.Interface, clusterName string
...
+			return err
 		}
-		controller := batchv1beta1.CronJob{}
-		err = json.Unmarshal(str, &controller)
+		finalDoc := make(map[string]interface{})
+		finalDoc["metadata"] = yamlNode["metadata"]
+		finalDoc["apiVersion"] = "v1"
+		finalDoc["kind"] = "Pod"
+		finalDoc["spec"] = getPodSpec(yamlNode)
+		marshelledYaml, err := yaml.Marshal(finalDoc)
```
rbren: `s/mareshell/marshal` (below too)

pkg/kube/resources.go
```diff
@@ -168,222 +137,77 @@ func CreateResourceProviderFromAPI(kube kubernetes.Interface, clusterName string
...
-			return nil, err
+			logrus.Errorf("Invalid YAML: %s", string(contents))
+			return err
 		}
-		controller := batchv1beta1.CronJob{}
-		err = json.Unmarshal(str, &controller)
+		finalDoc := make(map[string]interface{})
+		finalDoc["metadata"] = yamlNode["metadata"]
+		finalDoc["apiVersion"] = "v1"
+		finalDoc["kind"] = "Pod"
```
rbren: I think we're losing the `Kind` from the file here. E.g. if the user has a `Deployment` manifest in their repo, we want that to show up as a `Deployment` in the dashboard.

pkg/validator/fullaudit_test.go
```diff
@@ -33,8 +33,8 @@ func TestGetTemplateData(t *testing.T) {
 
 	sum := CountSummary{
 		Successes: uint(0),
-		Warnings:  uint(9),
-		Errors:    uint(9),
+		Warnings:  uint(1),
```
rbren: Looks like we lost a lot of tests - why is that?

pkg/config/supportedcontrollers.go
```diff
@@ -10,7 +10,7 @@ import (
 	appsv1beta1 "k8s.io/api/apps/v1beta1"
 	appsv1beta2 "k8s.io/api/apps/v1beta2"
 	batchv1 "k8s.io/api/batch/v1"
-	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	batchV1beta1 "k8s.io/api/batch/v1beta1"
```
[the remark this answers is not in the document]
baderbuddy: This was actually a mistake, I don't know how I accidentally deleted this at some point and was trying to put it back in, I'll revert it back to lowercase.

baderbuddy:
I fixed up all of the tests, a few don't have as much data and I added TODOs there, that's because we're only retrieving things at the pod level now and all of the fake controllers haven't added pods.

pkg/validator/controllers/generic.go
```diff
@@ -0,0 +1,126 @@
...
+	return g.Name
+}
+
+// GetNamespace is inherited by all controllers using generic controller to get the namespace of the controller
+func (g GenericController) GetNamespace() string {
+	return g.Namespace
+}
+
+// LoadControllers loads a list of controllers from the kubeResources Pods
+func LoadControllers(pods []kubeAPICoreV1.Pod, dynamicClientPointer *dynamic.Interface, restMapperPointer *meta.RESTMapper) []GenericController {
```
rbren: Can we move this logic into `pkg/kube`? I'd like to keep all the API interactions in there.

pkg/validator/controllers/generic.go
```diff
@@ -0,0 +1,126 @@
...
+	controller.KindString = "Pod"
+	controller.CreatedTime = controller.GetObjectMeta().CreationTimestamp.Time
+
+	owners := controller.GetObjectMeta().OwnerReferences
+	if dynamicClientPointer == nil || restMapperPointer == nil {
+		return controller
+	}
+	// If an owner exists then set the name to the controller.
+	// This allows us to handle CRDs creating Controllers or DeploymentConfigs in OpenShift.
+	for len(owners) > 0 {
```
rbren: Should we log an info message if there are multiple owners?

pkg/validator/controllers/generic.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// GenericController is a base implementation with some free methods for inherited structs
+type GenericController struct {
+	Name        string
+	Namespace   string
+	PodSpec     kubeAPICoreV1.PodSpec
+	ObjectMeta  kubeAPIMetaV1.ObjectMeta
+	Kind        config.SupportedController
+	KindString  string
```
rbren: I don't like the idea of tracking both `Kind` and `KindString` - it's not immediately apparent why they'd diverge (though I can see from the code that they do)

Is there anything wrong with always setting `Kind` to the controller Kind, and getting rid of the `SupportedController` type?

pkg/validator/controllers/cronjob.go
```diff
@@ -2,42 +2,20 @@ package controllers
...
-	return c.K8SResource.ObjectMeta
-}
-
 // NewCronJobController builds a new controller interface for Deployments
-func NewCronJobController(originalDeploymentResource kubeAPIBatchV1beta1.CronJob) Interface {
-	controller := CronJobController{}
-	controller.Name = originalDeploymentResource.Name
-	controller.Namespace = originalDeploymentResource.Namespace
-	controller.K8SResource = originalDeploymentResource
+func NewCronJobController(originalResource kubeAPIBatchV1beta1.CronJob) GenericController {
```
rbren: Seems like these functions are just used in the tests - possible to nuke them or move them to the `/test` package?

pkg/validator/output.go
```diff
@@ -62,11 +59,12 @@ type ResultSet map[string]ResultMessage
...
-	Namespace string
-	Kind      string
-	Results   ResultSet
-	PodResult PodResult
+	Name        string
+	Namespace   string
+	Kind        string
+	Results     ResultSet
+	PodResult   PodResult
+	CreatedTime time.Time
```
rbren: Why did this get added?

baderbuddy: The time was there specifically for the dedupe logic

pkg/validator/controllers/cronjob.go
baderbuddy: The validator webhook still uses these too

pkg/validator/controllers/generic.go
baderbuddy: Got the logging in there now.

pkg/validator/controllers/generic.go
baderbuddy: I got rid of the Kind on the GenericController, but left the SupportedController type because it still seemed useful for a few things, especially it seems like we use it on the validation webhook side.

PR #241, opened 2020-02-06
Refactor commands to use Cobra

baderbuddy:
Changed the interface to use Cobra commands.


cmd/polaris/root.go
```diff
@@ -0,0 +1,90 @@
...
+
+func init() {
+	// Flags
+	rootCmd.PersistentFlags().StringVarP(&configPath, "config", "c", "", "Location of Polaris configuration file.")
+	rootCmd.PersistentFlags().BoolVarP(&disallowExemptions, "disallow-exemptions", "", false, "Disallow any exemptions from configuration file.")
+	rootCmd.PersistentFlags().StringVarP(&logLevel, "log-level", "", logrus.InfoLevel.String(), "Logrus log level.")
+	flag.Parse()
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+}
+var c conf.Configuration
```
rbren: Since this is global now, can we give it a better name?

cmd/polaris/version.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func init() {
+	rootCmd.AddCommand(versionCmd)
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Prints the current version of the tool.",
+	Long:  `Prints the current version.`,
```
rbren: I think you've got short/long reversed :)

docs/usage.md
```diff
@@ -119,8 +119,8 @@ kubectl apply -f https://github.com/fairwindsops/polaris/releases/latest/downloa
 
 ### Helm
 ```bash
-helm repo add reactiveops-stable https://charts.reactiveops.com/stable
-helm upgrade --install polaris reactiveops-stable/polaris --namespace polaris \
+helm repo add fairwindsops-stable https://charts.fairwinds.com/stable
```
rbren: `s/fairwindsops/fairwinds`, below too

main.go
```diff
@@ -15,279 +15,17 @@
...
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
-	"sigs.k8s.io/yaml"
+	"github.com/fairwindsops/polaris/cmd/polaris"
 )
 
 const (
 	// Version represents the current release version of Polaris
 	Version = "0.6.0"
+	// TODO make sure this gets set.
+	commit = "n/a"
```
rbren: What's this for?

baderbuddy: Copied from Goldilocks and not needed and can be safely removed

baderbuddy: Well, it was used for the version output, maybe it still would be valuable

baderbuddy:
Tests are passing and ready for another look

main.go
rbren: I'd say let's get rid of it. There's no way to check the commit hash in without changing the commit hash, so I can't imagine what this is meant for.

Issue #244, opened 2020-02-25
Webhook does not cover all possible versions of deployments


ingshtrom:
### Installation Process
I manually installed the dashboard and webhook into an EKS cluster via the [instructions here](https://github.com/FairwindsOps/polaris/blob/master/docs/usage.md#dashboard).

Nothing was modified, except increasing the log level on the webhook deployment.

### Polaris Version
```
0.6
```

### Expected Behavior
My deployment gets restricted

### Actual Behavior
I can deploy my test application without issues.

### Steps to Reproduce
1. `kubectl apply -f file.yaml` YAML from below.

Alternatively, if the YAML is modified so that it reads `apisVersion: apps/v1`, then `kubectl apply -f ...` fails as expected. 

### Additional Context
Here is the YAML I'm deploying to my cluster via `kubectl apply -f file.yaml`
```
---

apiVersion: apps/v1beta2
kind: Deployment
metadata:
  name: router-test
  namespace: infra-system
spec:
  selector:
    matchLabels:
      app: router-test
  replicas: 1
  template:
    metadata:
      labels:
        app: router-test
    spec:
      restartPolicy: Always
      containers:
        - name: web
          image: nginx
          imagePullPolicy: Always
          livenessProbe:
            httpGet:
              path: /
              port: 80
          ports:
            - containerPort: 80

```

I also inspected the ValidationWebhookConfiguration that is generated the the `polaris-webhook` container, and it seems that it does only bind against the `apps/v1` apis for `deployment` objects. I haven't written a webhook validator before, but my _hunch_ is that is where the problem lies. Honestly, I'm curious why Kubernetes doesn't deprecate the old versions of the API in the first place :man_shrugging:.

It would be cool if as an operator I could restrict the APIs that people using a cluster could use via the Polaris tool, but I realize that likely adds a lot of complexity in re-generating the `ValidationWebhookConfiguration` every time the configmap is modified... or maybe we just need _restart_ the webhook pod and we're good? :thinking: 

### References
n/a

rbren:
Thanks for pointing this out!

We've got a fix above in flight. It will likely go out with the 1.0 release.

PR #245, opened 2020-02-25
Add support for more controller versions

rbren:
Fixes https://github.com/FairwindsOps/polaris/issues/244

The most disruptive change here is that if building the webhook fails, we don't exit the process. This way if we try to watch a type/version that's not available in the cluster, it won't fail to start.


pkg/kube/resources.go
```diff
@@ -121,52 +122,49 @@ func CreateResourceProviderFromAPI(kube kubernetes.Interface, clusterName string
...
 		return nil, err
 	}
-	daemonSets, err := kube.AppsV1().DaemonSets("").List(listOpts)
+	cronJobs, err := getCronJobs(kube)
 	if err != nil {
-		logrus.Errorf("Error fetching DaemonSets %v", err)
 		return nil, err
 	}
-	jobs, err := kube.BatchV1().Jobs("").List(listOpts)
+	daemonSets, err := kube.AppsV1().DaemonSets("").List(listOpts)
```
makoscafee: could this use `getDaemonSets` too? I think you will still miss `appsv1beta2`

rbren: Good catch! I added some tests to make sure we don't miss these.

PR #247, opened 2020-03-06
Command argument error `polaris --webhook`

BobVanB:


rbren:
Good catch! Opened a second PR to fix the dashboard yaml as well.

PR #256, opened 2020-03-14
Fix chart repository name in docs

arlyon:
charts.fairwindsops.com doesn't resolve, fixed according to https://github.com/FairwindsOps/charts


rbren:
Thanks for the fix!

PR #270, opened 2020-03-30
Delete controllers package

rbren:


pkg/kube/workload.go
```diff
@@ -0,0 +1,65 @@
...
+	ObjectMeta  kubeAPIMetaV1.ObjectMeta
+	Kind        string
+	CreatedTime time.Time
+}
+
+// NewGenericWorkload builds a new workload for a given Pod
+func NewGenericWorkload(originalResource kubeAPICoreV1.Pod, dynamicClientPointer *dynamic.Interface, restMapperPointer *meta.RESTMapper) GenericWorkload {
+	workload := GenericWorkload{}
+	workload.Name = originalResource.Name
+	workload.Namespace = originalResource.Namespace
```
rbren: @baderbuddy I can't figure out why `originalResource.Name` is defined, [based on the godoc](https://godoc.org/k8s.io/api/core/v1#Pod)

Looks like we're tracking `Name` and `Namespace` in addition to `ObjectMeta` - any thoughts on getting rid of one or the other?

baderbuddy: Yeah using ObjectMeta to get at Name and Namespace makes sense. I'm pretty sure the `.Name` and `.Namespace` are just shorthand.

examples/config-full.yaml
```diff
@@ -72,13 +72,6 @@ customChecks:
           not:
             pattern: ^quay.io
 
-controllersToScan:
```
baderbuddy, line 75: Do we want to leave this to limit the validating webhook?

pkg/kube/workload.go
baderbuddy: Ohh I misunderstood your question. Ooh that's a great point. Right now we keep the Pod's Meta no matter what, but the Name matches the top level controller. I could see value in the metadata especially for Annotations, but for the most part I think just the name and Namespace are all we really need.

baderbuddy (approved):
This looks so much cleaner. I like it!

PR #285, opened 2020-05-11
Add the capability for controller level checks

baderbuddy:
Brings along the JSON of the top level object. I also added a single check mostly for documentation purposes (but it could be a pretty useful check too)


pkg/validator/schema.go
```diff
@@ -124,6 +126,17 @@ func applyPodSchemaChecks(conf *config.Configuration, controller kube.GenericWor
 		if err != nil {
 			return nil, err
 		} else if check == nil {
+			check, err = resolveCheck(conf, checkID, controller, config.TargetController, false)
```
rbren, line 129: I think this should go in its own function, `applyControllerSchemaChecks`, which could get called in `pkg/validator/controller.go`. That's similar to how we do both pod and container targets.

baderbuddy: Ohh yeah that makes a lot of sense, I was thinking of it still as controller checks, but it's still explicitly not a podSchemaCheck.

PR #291, opened 2020-05-18
don't push docker images on forked PRs

rbren:
This should fix the issue with dependabot

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ references:
 
   docker_build: &docker_build
     run:
-      name: Docker login, build, and push
+      name: Docker login and build
       command: |
         docker-pull -f .circleci/build.config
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -89,18 +89,23 @@ references:
         echo "8dbad6683d6fc9367e637e6eed8e01a0d63c9660  goreleaser.deb" | sha1sum -c
         sudo dpkg -i goreleaser.deb
         rm goreleaser.deb
+
+  docker_build: &docker_build
+    run:
+      name: Docker login, build, and push
```
jordandoig, line 95: Could we update this name? Not sure if it has any implications outside of clarity, but I think that's reason enough for a simple change

rbren: Good catch!

PR #298, opened 2020-05-18
Pack config.yaml in Docker

rbren:


rbren:
I decided to remove the exemptions from `config-full.yaml`, since they're constantly falling out of date.

`config-full.yaml` is really an example usage, but probably wouldn't be dropped in for anyone. `config.yaml` is the default though.

PR #300, opened 2020-05-18
Release 1.0.0

rbren:


mambax:
Stupid question from my side. With 1.0 the CLI 'commander' is changed to cobra we have seen (and with that how you hand in arguments afaict). Did that find a place in the new docs? Not for me I've seen it now, but we firstly struggled what changed and then finding the commit hiding it :)

rbren:
Good question! Yes everything is documented in [usage](/docs/usage.md), but I also added a note to the CHANGELOG. Thanks for pointing that out!

rbren:
FYI - we're just waiting on the [Quay outage](https://status.quay.io/)

mambax:
nervous:tenor

Issue #313, opened 2020-06-01
Navigation logo redirection doesn't obey basePath


yiwei-C:
### Installation Process
Installed with Helm Chart on Kubernetes.

### Polaris Version
```
1.0.2
```

### Expected Behavior
While running Polaris with basePath, clinking the Polaris icon on the navigation bar should redirect back to the main page.

### Actual Behavior
It doesn't support basePath and redirects to "/".

### Steps to Reproduce
- Run Polaris with "--base-path".
- Clink the Polaris icon.

### Additional Context
Here is the link to this bug: https://github.com/FairwindsOps/polaris/blob/2b15f11d57d9d497e7f96dd42f0befd84a5998ff/pkg/dashboard/templates/navbar.gohtml#L4

rbren:
Fixed in 1.0.3 - thanks!

PR #319, opened 2020-06-01
fix basepath link

rbren:


rbren:
Fixes https://github.com/FairwindsOps/polaris/issues/313

PR #322, opened 2020-06-03
Bump deployments to polaris:1.0.2

ajayk:


rbren:
Thanks for reaching out!

The `1` tag actually refers to `1.0.2` - you can see [in the tags](https://quay.io/repository/fairwinds/polaris?tab=tags) that they point to the same SHA. When we release `1.0.3`, the `1` and `1.0` tags will be updated to point to `1.0.3`

But we actually intended to pin the _minor_ version here, not the _major_ version. So I've opened up https://github.com/FairwindsOps/polaris/pull/323 to address this

PR #326, opened 2020-06-03
fix goreleaser

rbren:


.circleci/config.yml
```diff
@@ -168,6 +168,7 @@ jobs:
       - *set_environment_variables
       - *install_goreleaser
       - run: go get -u github.com/gobuffalo/packr/v2/packr2
+      - run: git checkout -- . # FIXME: the go get makes the directory dirty
```
baderbuddy, line 171: Is this just to reset the directory? Would a git reset work instead? I'm mostly confused by the `-- .` syntax, don't think I understand what's going on there.

rbren: This is how I always tell git "undo all the changes I've made". What do you do?

jordandoig: I had to google your syntax, Robert. Personally, I always do `git reset --hard` to clear any changes to tracked files, does your syntax get rid if changes to new/untracked files?

rbren: no untracked files are unchanged. I use `git clean -df path/` to clean those up

I always thought `git reset --hard` changed revision history, so I avoided it. It's definitely a clearer syntax though

Issue #328, opened 2020-06-05
PodResult present in auditing a `Service`


amitsaha:
### Installation Process

CLI mode, using locally.

### Polaris Version
```
Polaris version:1.0.3

```

### Expected Behavior

I don't expect to see the `PodResult` reported with any issues.

### Actual Behavior

Issues are reported.

### Steps to Reproduce

Consider the following `Service` definition:

```
apiVersion: v1
kind: Service
metadata:
  name: http-echo
spec:
  ports:
  - port: 5678
    protocol: TCP
    targetPort: 5678
  selector:
    app: http-echo

```

When I run `polaris audit`, I get this:

```
$ polaris audit --audit-path test-data/base-valid.yaml  
{
  "PolarisOutputVersion": "1.0",
  "AuditTime": "0001-01-01T00:00:00Z",
  "SourceType": "Path",
  "SourceName": "test-data/base-valid.yaml",
  "DisplayName": "test-data/base-valid.yaml",
  "ClusterInfo": {
    "Version": "unknown",
    "Nodes": 0,
    "Pods": 1,
    "Namespaces": 0,
    "Controllers": 1
  },
  "Results": [
    {
      "Name": "http-echo",
      "Namespace": "",
      "Kind": "Service",
      "Results": {},
      "PodResult": {
        "Name": "",
        "Results": {
          "hostIPCSet": {
            "ID": "hostIPCSet",
            "Message": "Host IPC is not configured",
            "Success": true,
            "Severity": "danger",
            "Category": "Security"
          },
          "hostNetworkSet": {
            "ID": "hostNetworkSet",
            "Message": "Host network is not configured",
            "Success": true,
            "Severity": "warning",
            "Category": "Networking"
          },
          "hostPIDSet": {
            "ID": "hostPIDSet",
            "Message": "Host PID is not configured",
            "Success": true,
            "Severity": "danger",
            "Category": "Security"
          }
        },
        "ContainerResults": []
      },
      "CreatedTime": "0001-01-01T00:00:00Z"
    }
  ]
}```

### Additional Context

### References

baderbuddy:
Thanks for reporting this. I think I see what the issue is, it should be a pretty easy fix.

PR #329, opened 2020-06-05
Don't assume all objects have pods.

baderbuddy:
The string based audits are picking up services and other objects as controllers. See issue #328

### DIFF
--- a/pkg/kube/resources.go
+++ b/pkg/kube/resources.go
@@ -228,7 +228,10 @@ func GetPodSpec(yaml map[string]interface{}) interface{} {
 			return GetPodSpec(childYaml.(map[string]interface{}))
 		}
 	}
-	return yaml
+	if _, ok := yaml["containers"]; ok {
+		return yaml
+	}
+	return nil
 }
 
 func addResourceFromString(contents string, resources *ResourceProvider) error {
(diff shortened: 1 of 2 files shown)


rbren:
Oops!

Mind adding a quick test case?

baderbuddy:
Looks like we actually already had a test, we just set the wrong values for it, whoops!

pkg/kube/resources_test.go
```diff
@@ -44,7 +44,7 @@ func TestGetMultipleResourceFromSingleFile(t *testing.T) {
 
 	assert.Equal(t, 0, len(resources.Nodes), "Should not have any nodes")
 
-	assert.Equal(t, 4, len(resources.Controllers), "Should have four controllers")
+	assert.Equal(t, 1, len(resources.Controllers), "Should have four controllers")
```
rbren, line 47: ...why did this change? Also message is off now.

rbren: Ahh I see the reason now. Thanks!

baderbuddy: It was actually wrong before apparently. 3 of those "Controllers" were empty yaml documents.

PR #360, opened 2020-06-22
throw error when severity isn't set for custom check

rbren:


samuelb:
I would rather show a warning instead of error and return. This would allow to comment out a custom check without the need to remove/comment out the whole custom check definition as well. Only for the convince.

Issue #351, opened 2020-06-16
Query Object Metadata


k45c4d3:
Linking closed issue https://github.com/FairwindsOps/polaris/issues/338

@rbren 

Sorry for the delay getting back to you on this.

To clarify right now I'm primarily interested in the `spec.template.metadata.labels`

So far I haven't gotten it to work. I took your exact config from above and used it as the config file to run an audit on the sample deployment manifest in the original comment. This is how I have been testing all of my checks so far. I get success no matter what the label is.

`polaris audit -c ~/Desktop/polarisconf.yaml --audit-path ~/Desktop/polaristestchecks.yaml --log-level debug`

```spec:
  replicas: 1
  selector:
    matchLabels:
      app: label-test
  strategy: {}
  template:
    metadata:
      labels:
        app: label-test
```
```
  "Results": [
    {
      "Name": "polaris-test-deploy",
      "Namespace": "",
      "Kind": "Deployment",
      "Results": {},
      "PodResult": {
        "Name": "",
        "Results": {
          "podNameLabelExists": {
            "ID": "podNameLabelExists",
            "Message": "label is correct",
            "Success": true,
            "Severity": "warning",
            "Category": ""
```
```
spec:
  replicas: 1
  selector:
    matchLabels:
      app: label-test
  strategy: {}
  template:
    metadata:
      labels:
        app: INCORRECT_LABEL
    spec:
      containers:
      - image: centos:7
        name: centos
        resources: {}
```
```
 "Results": [
    {
      "Name": "polaris-test-deploy",
      "Namespace": "",
      "Kind": "Deployment",
      "Results": {},
      "PodResult": {
        "Name": "",
        "Results": {
          "podNameLabelExists": {
            "ID": "podNameLabelExists",
            "Message": "label is correct",
            "Success": true,
            "Severity": "warning",
            "Category": ""
```

samuelb:
Looks like the checks with `target: Pod` only looks at `spec.template.spec` and below.

eg. when looking at this Deployment spec:
```
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    metadata:
      labels:
        app: INCORRECT_LABEL
    spec:
      containers:
      - image: centos:7
        name: centos
        resources: {}
```
This customChecks rule would be successfully:
```
testCustomChecks:
  successMessage: All OK
  failureMessage: Nothing is OK
  target: Pod
  schema:
  '$schema': http://json-schema.org/draft-07/schema
  required:
  - containers
[...]

PR #364, opened 2020-06-26
Show controller check results on dashboard

samuelb:
As I mentioned in https://github.com/FairwindsOps/polaris/issues/351,
custom check results for `target: Controller` have not been shown on the
Dashboard.


rbren:
Good catch! Thanks!

Issue #327, opened 2020-06-04
Ignore orphaned pods


irasnyd:
### Installation Process
 

EKS cluster, version 1.15, using helm.

Command line:
```bash
$ helm upgrade --install polaris fairwinds-stable/polaris --version 1.0.2 --namespace kube-system -f polaris-values.yaml
```

`polaris-values.yaml` contents:
```yaml
# Override the tag provided in the chart as the unconstrained "1.0"
# to be the more constrained "1.0.3" (a specific release).
image:
  tag: "1.0.3"

# Enable the dashboard over HTTP for internal users
dashboard:
  enable: true
  ingress:
    enabled: true
    annotations:
      kubernetes.io/ingress.class: nginx-ingress-private
    hosts:
      - "polaris.internal.domain"
```

### Polaris Version

Version 1.0.3 Docker Image

```
$ k -n kube-system get deploy polaris-dashboard -o wide
NAME                READY   UP-TO-DATE   AVAILABLE   AGE   CONTAINERS   IMAGES                            SELECTOR
polaris-dashboard   1/1     1            1           15d   dashboard    quay.io/fairwinds/polaris:1.0.3   app=polaris,app.kubernetes.io/instance=polaris,app.kubernetes.io/name=polaris,component=dashboard

```

### Expected Behavior

The dashboard should be working.

### Actual Behavior

Webpage with contents:

```
Error fetching Kubernetes resources
```

Logs from container:
```
$ k -n kube-system logs polaris-dashboard-dcb6c8b9c-9bpkw 
time="2020-06-04T17:36:42Z" level=info msg="Starting Polaris dashboard server on port 8080"
time="2020-06-04T17:37:04Z" level=error msg="Cache missed ReplicaSet/prod/telbot-gunicorn-7844c9db9d again"
time="2020-06-04T17:37:04Z" level=error msg="Error loading controllers from pods: Could not retrieve parent object"
time="2020-06-04T17:37:04Z" level=error msg="Error fetching Kubernetes resources Could not retrieve parent object"
```

CURL output of same request:
```
$ curl -v -L 'http://polaris.internal.domain/'
* About to connect() to polaris.internal.domain port 80 (#0)
*   Trying 10.100.25.113...
* Connected to polaris.internal.domain (10.100.25.113) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.29.0
> Host: polaris.internal.domain
> Accept: */*
> 
[...]

PR #370, opened 2020-07-08
Fix error on Controller checks on orphaned pods

samuelb:
Related issue: https://github.com/FairwindsOps/polaris/issues/327

Polaris errored when it wanted to run a Contoller check on a Pod which is
termination state and the related Controller is gone already. The
Dashboard showed `Error running audit`, the logs:
```
WARN[0014] An error occured validating controller:error parsing JSON bytes: unexpected end of JSON input
ERRO[0014] Error getting audit data: error parsing JSON bytes: unexpected end of JSON input
```

As this should happen only when a Deployment/StatefulSet/DaemonSet was
deleted, I consider it safe to ingore such problem on orphaned pods.

With this patch, Polaris will log a warning and continue instead.

### DIFF
--- a/pkg/validator/schema.go
+++ b/pkg/validator/schema.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/fairwindsops/polaris/pkg/config"
 	"github.com/fairwindsops/polaris/pkg/kube"
+	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rbren:
Thanks for the PR! I'm trying to repro the original issue to verify, but having a bit of trouble. Will update on #327

samuelb:
I'm closing this in favour of #371

PR #378, opened 2020-07-21
Update config.yaml

rbren:


baderbuddy:
I was curious if we had documentation for our users in case anyone else was confused about why to set image pull policy to always. I found this. https://github.com/FairwindsOps/polaris/blob/master/docs/check-documentation/images.md

It explicitly states that we don't enable this by default. I still feel like it should be enabled but maybe we should update that doc too.

PR #400, opened 2020-08-26
Update dependencies

baderbuddy:
For right now the test folder is incomplete and the deployment yaml. All of the actual code changes should be ready for review though.


makoscafee (approved):
LGTM. just had few question

cmd/polaris/webhook.go
```diff
@@ -74,79 +42,34 @@ var webhookCmd = &cobra.Command{
...
-			logrus.Debugf("Error reading namespace information: %v", err)
-		}
-
-		polarisNamespace := string(polarisNamespaceBytes)
-		if polarisNamespace == "" {
-			polarisNamespace = polarisResourceName
-			logrus.Debugf("Could not determine current namespace, creating resources in %s namespace", polarisNamespace)
+		_, err = os.Stat("/opt/cert/tls.crt")
+		if os.IsNotExist(err) {
+			time.Sleep(time.Second * 10)
```
rbren, line 57: why sleep before `panic`ing?

cmd/polaris/webhook.go
```diff
@@ -74,79 +42,34 @@ var webhookCmd = &cobra.Command{
...
+			panic("Cert does not exist")
 		}
+		server := mgr.GetWebhookServer()
+		server.CertName = "tls.crt"
+		server.KeyName = "tls.key"
 
 		logrus.Info("Setting up webhook server")
```
rbren, line 64: I think these log messages might need some thought

cmd/polaris/webhook.go
baderbuddy: Oh I forgot about that. It was to cut down on the chances of a crash loop backoff and give the job more time to update the cert.

baderbuddy:
The ValidatingWebhookConfiguration explicitly calls out which types of objects to watch for. Polaris itself doesn't really care because it's treating them all the same anyways.

rbren:
OK yeah I see that in the chart now - but how does that get plumbed in here?

.circleci/config.yml
```diff
@@ -27,24 +27,24 @@ references:
...
-        curl -sLO https://storage.googleapis.com/kubernetes-release/release/v1.12.7/bin/linux/amd64/kubectl
+        curl -sLO https://storage.googleapis.com/kubernetes-release/release/v1.18.6/bin/linux/amd64/kubectl
         chmod 0755 kubectl
         sudo mv kubectl /usr/local/bin/
         kubectl version --client
 
 
         echo "Creating Kubernetes Cluster with Kind"
-        kind create cluster --wait=90s
+        kind create cluster --wait=90s --image kindest/node:v1.15.11
```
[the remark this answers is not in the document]
baderbuddy: It should work fine with other versions, but 1.16 removes a lot of deprecated APIs so we'd lose a lot in the testing if we test with 1.16 or higher.

baderbuddy:
Either the helm install or the webhook deploy yaml. The Polaris code itself is totally agnostic.

cmd/polaris/webhook.go
rbren: Mind taking a look at these? It looks like this one now comes _after_ we set up the webhook server. The message below about `listening on port` seems out of place too

baderbuddy: Yeah, I ripped out a lot of the log messages (I don't think they really added value), I kept the one about listening on port x, right at the end.

rbren (approved):

PR #421, opened 2020-09-28
Implement namespace support for exceptions

mblaschke:
Add namespace support for exceptions for shared clusters.

### DIFF
--- a/pkg/config/exemptions.go
+++ b/pkg/config/exemptions.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"fmt"
 	"strings"
 )
 
@@ -16,7 +15,6 @@ func (conf Configuration) IsActionable(ruleID, namespace, controllerName string)
 
 	for _, example := range conf.Exemptions {
 		if example.Namespace != "" && example.Namespace != namespace {
-			fmt.Println("not matches: " + example.Namespace + " -> " + namespace)
 			continue
 		}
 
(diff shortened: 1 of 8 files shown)


pkg/config/exemptions.go
```diff
@@ -14,10 +15,16 @@ func (conf Configuration) IsActionable(ruleID, controllerName string) bool {
 	}
 
 	for _, example := range conf.Exemptions {
+		if example.Namespace != "" && example.Namespace != namespace {
+			fmt.Println("not matches: " + example.Namespace + " -> " + namespace)
```
rbren, line 19: I think we can get rid of this `Println`

rbren (approved):
Great work, and thanks for adding a test! Just one little note

rbren:
Can you also add documentation here: https://github.com/FairwindsOps/polaris/blob/master/docs/usage.md#exemptions

pkg/config/exemptions.go
mblaschke: ah, sure.. fixed :)

mblaschke:
Done

marcemv90:
With this new feature is it possible to ignore completely a namespace? Because I would like to get rid of the kube-system namespace in the dashboard.

baderbuddy (approved):
This looks really good to me! Thank you!

baderbuddy:
@marcemv90 this should allow all controllers within a namespace to be exempted (since it matches on prefix you could add a controller-name of `""`) but you'd have to specify the rules to exempt still

PR #438, opened 2020-11-02
Dashboard with exceptions as ConfigMap

just1689:
Dashboard with exceptions as ConfigMap

### DIFF
--- a/deploy/dashboard-exceptions.yaml
+++ b/deploy/dashboard-exceptions.yaml
@@ -1,0 +1,367 @@
+---
+# Source: polaris/templates/0-namespace.yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: polaris
+---
+# Source: polaris/templates/rbac.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: polaris
+  namespace: polaris
+  labels:
+    app: polaris
+---
+# Source: polaris/templates/rbac.yaml
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: polaris
+  labels:
+    app: polaris
+rules:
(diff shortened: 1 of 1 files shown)


rbren:
Hi @just1689! Thanks for opening this.

A couple issues here:
* these deploy files are autogenerated from our helm chart. Any changes here should be added as a command in [the script](https://github.com/FairwindsOps/polaris/blob/master/scripts/generate-deployment-files.sh)
* more importantly, these exemptions should already be included in the default installation of polaris - if no configmap is provided, it should use `examples/config.yaml`. Is that not the behavior you're seeing?

just1689:
Howdy

Ah I see I have misunderstood the origin of those files. Helm as the source makes sense.

I'll jot out my though process. This could help you think through the usability of the repo. Think of it as thirdparty feedback. This is roughly path I took that led me here:

- I would like try the project
- Spun up the Quickstart
- Wanted to add exemptions for a few different things in my cluster. Googled "Polaris exceptions". First result: https://github.com/FairwindsOps/polaris/blob/master/docs/usage.md
- Saw the bit about `exemptions`. That's what I want. Explored the repo and found the default. Figured I would start with that.
- Wanted explicit, declarative YAML on disk so I copied the Quickstart YAML and modified it to add a ConfigMap.
- It worked out well and I went on to customize it.
- Later thought that other people might want to do the exact same thing - that is use a ConfigMap for the config.
- Overwrote my ConfigMap yaml with the examples/config.yaml
- Sent through pull request

I think the questions that follow include:
Is this a common requirement? Will others want to spin it up and modify the config, tare down and change? Will they prefer to work with YAML on disk (like me) or are they going to work with the Helm chart?

The answers to those questions will maybe help you determine if this is a valid addition to the quick start - or something similar.

One idea is that the Quickstart could include an example of how to spin it up with a custom config?

PR #450, opened 2020-12-14
Bump version to 2.0.0

rbren:


baderbuddy (changes requested):
Since the controller-runtime stuff is in, I think we'd have to do the 2.0 release instead of doing another 1.x release.

rbren:
Oh good call, thanks!

baderbuddy (approved):
Awesome! If you undo the changes to the webhook yaml (I had already freshly generated that with the 2.0 changes to the chart).

PR #454, opened 2020-12-17
change how controller checks are handled

rbren:
This fixes an inconsistency in how controller-level checks are handled between IaC and in-cluster resources.

The in-cluster resources had an `Object` wrapper around the JSON - this removes that.

This will need a major version bump.

Still working on testing this properly - our k8s mocks aren't quite working right.


baderbuddy (approved):
Ahh nice. It's really a bugfix so it might be worth doing in a minor version

rbren:
The issue is that someone might have written a custom check that assumes the top-level `Object`, which would break with the change :/

PR #455, opened 2020-12-17
Fix test fixtures, add a test for controllers

rbren:


pkg/kube/resources_test.go
```diff
@@ -27,12 +27,12 @@ func TestGetResourcesFromPath(t *testing.T) {
...
 	assert.Equal(t, "two", resources.Namespaces[0].ObjectMeta.Name)
 
-	assert.Equal(t, 8, len(resources.Controllers), "Should have eight controllers")
+	assert.Equal(t, 9, len(resources.Controllers), "Should have eight controllers")
 	namespaceCount := map[string]int{}
 	for _, controller := range resources.Controllers {
 		namespaceCount[controller.ObjectMeta.GetNamespace()]++
 	}
-	assert.Equal(t, 7, namespaceCount[""], "Should have seven controller in default namespace")
+	assert.Equal(t, 8, namespaceCount[""], "Should have seven controller in default namespace")
```
baderbuddy, line 35: Maybe we should get rid of these messages, both of them are out of sync now.

pkg/kube/resources_test.go
```diff
@@ -99,7 +96,19 @@ func TestGetResourceFromAPI(t *testing.T) {
...
+
+	expectedNames := map[string]bool{
+		"deploy":      false,
+		"job":         false,
+		"cronjob":     false,
+		"statefulset": false,
+		"daemonset":   false,
+	}
+	for _, ctrl := range resources.Controllers {
+		expectedNames[ctrl.ObjectMeta.GetName()] = true
```
baderbuddy, line 109: I like this way of handling. I wonder if the assert library we use has something built in already or if we could contribute to it

baderbuddy (approved):
This looks great to me. There were a couple messages in the tests that were a little inaccurate and that's it.

pkg/kube/resources_test.go
rbren: I think there's something like it, but it expects the same object type

PR #457, opened 2020-12-18
Add to exemptions docs

baderbuddy:
Added a little more to the exemption docs to talk about how to exempt all controllers and all rules.

### DIFF
--- a/docs-md/customization/exemptions.md
+++ b/docs-md/customization/exemptions.md
@@ -18,19 +18,24 @@ kubectl annotate deployment my-deployment polaris.fairwinds.com/cpuRequestsMissi
 
 ## Config
 
-To exempt a controller via the config, you have to specify a namespace (optional), a list of controller names and a list of rules, e.g.
+To exempt a controller via the config, you have to specify a namespace (optional), a list of controller names and a list of rules (optional), e.g.
 ```yaml
 exemptions:
   # exemption valid for kube-system namespace
   - namespace: kube-system
     controllerNames:
-      - dns-controller
+    - dns-controller
     rules:
-      - hostNetworkSet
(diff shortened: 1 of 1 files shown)


makoscafee (approved):

docs-md/customization/exemptions.md
```diff
@@ -18,19 +18,24 @@ kubectl annotate deployment my-deployment polaris.fairwinds.com/cpuRequestsMissi
 
 ## Config
 
-To exempt a controller via the config, you have to specify a namespace (optional), a list of controller names and a list of rules, e.g.
+To exempt a controller via the config, you have to specify a namespace (optional), a list of controller names and a list of rules (optional), e.g.
```
rbren, line 21:
```suggestion
To exempt a controller via the config, you must specify one or more controller names (nb: any controller with this _prefix_ will be exempted). You can also specify particular namespaces or a list of particular rules. For example:
```

docs-md/customization/exemptions.md
```diff
@@ -18,19 +18,24 @@ kubectl annotate deployment my-deployment polaris.fairwinds.com/cpuRequestsMissi
...
   # exemption valid in all namespaces
   - controllerNames:
-      - dns-controller
+    - dns-controller
     rules:
-      - hostNetworkSet
+    - hostNetworkSet
 ```
 
+If no rules are specified then every rule is exempted. Controller names are matched as a prefix, so an empty string will match every controller.
```
rbren, line 41: I would suggest moving this up above the snippet

baderbuddy: Good call, I want to test it out first, Dominik was saying it didn't work for him.

Issue #465, opened 2020-12-29
Admission controller: Namespace and controller information lost before validation


HansK-p:
### Installation Process
Polaris was installed using the Helm chart 3.0.0, with a docker image based on the master branch and with the following values.yaml:
<pre>
image:
  repository: k8sfaith/polaris
  tag: "2020-12-29"
  pullPolicy: Always

webhook:
  enable: true

config:
  checks:
    privilegeEscalationAllowed: danger

  exemptions:
    - namespace: ingress-nginx
      controllerNames:
        - ingress-nginx-controller
      containerNames:
        - controller
      rules:
        - privilegeEscalationAllowed
        - insecureCapabilities
        - notReadOnlyRootFilesystem

    - controllerNames:
        - ingress-nginx-controller
      containerNames:
        - controller
      rules:
        - privilegeEscalationAllowed
        - insecureCapabilities
        - notReadOnlyRootFilesystem

#    - containerNames:
#        - controller
#      rules:
#        - privilegeEscalationAllowed
#        - insecureCapabilities
#        - notReadOnlyRootFilesystem
</pre>

The deployment is installed using the following yaml file:
<pre>
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    name: ingress-nginx
  name: ingress-nginx
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: ingress-nginx
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: ingress-nginx
  replicas: 1
  template:
    metadata:
      labels:
        app.kubernetes.io/name: ingress-nginx
    spec:
      containers:
        - name: controller
          image: "k8s.gcr.io/ingress-nginx/controller:v0.42.0@sha256:f7187418c647af4a0039938b0ab36c2322ac3662d16be69f9cc178bfd25f7eee"
          imagePullPolicy: IfNotPresent
          securityContext:
            capabilities:
                drop:
                - ALL
                add:
                - NET_BIND_SERVICE
            runAsUser: 101
            allowPrivilegeEscalation: true
          livenessProbe:
            httpGet:
              path: /healthz
              port: 10254
              scheme: HTTP
          readinessProbe:
            httpGet:
              path: /healthz
              port: 10254
              scheme: HTTP
          resources: 
            requests:
              cpu: 100m
              memory: 90Mi
</pre>

### Polaris Version
[...]

PR #466, opened 2020-12-29
Added code to retain namespace and object name for webhook

HansK-p:
This code might solve #465, but it has only been tested lightly - so there is no guarantee.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -24,7 +24,6 @@ import (
 	"github.com/fairwindsops/polaris/pkg/config"
 	"github.com/fairwindsops/polaris/pkg/kube"
 	validator "github.com/fairwindsops/polaris/pkg/validator"
-
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


baderbuddy:
Thank you so much for this PR. I think this might work, but it doesn't feel as clean as making sure the `kube` package just sets the ObjectMeta correctly in the first place. We have the raw object there as a map, but it looks like we're incorrectly using the Pod instead of the controller. That'll mess with annotations too, so it's something that should get fixed.

HansK-p:
I'm pretty sure you are right and I'm just happy i the problem is solved as we are looking into using the Polaris admission controller.

We run our workloads on AKS and will start using Azure policies, but using the Polaris admission controller in addition has a value as Azure policies (currently) seems to block deployments only on pod level. So we are allowed to deploy the Deployment/Statefulset/Daemonset - but the Deployment/Statefulset/Daemonset is not allowed to create pods.

Being rejected already when doing the deployment gives a much better user experience + with Polaris we can use the same policy also in the pipeline.

rbren:
This was fixed by another PR. Thanks for pointing it out @HansK-p!

PR #468, opened 2021-01-04
Fix up zero-score issues

rbren:
This fixes a few issues when there are zero checks for a particular resource:
* the resource now gets a score of 100% instead of 0%
* the UI now shows green instead of red
* subsections will show `no checks available` instead of just being empty


baderbuddy (approved):
LGTM

PR #469, opened 2021-01-04
Add support for ingress objects

jordandoig:
This PR adds support for checking Ingress kubernetes objects


pkg/validator/schema.go
```diff
@@ -77,7 +80,7 @@ func parseCheck(rawBytes []byte) (config.SchemaCheck, error) {
 	}
 }
 
-func resolveCheck(conf *config.Configuration, checkID string, controller kube.GenericWorkload, container *corev1.Container, target config.TargetKind, isInitContainer bool) (*config.SchemaCheck, error) {
+func resolveCheck(conf *config.Configuration, checkID string, controller *kube.GenericWorkload, container *corev1.Container, ingress *v1beta1.Ingress, target config.TargetKind, isInitContainer bool) (*config.SchemaCheck, error) {
```
rbren, line 83: I think we can modify the function signature to make this a little more future proof:
```
resolveCheck(
  conf *config.Configuration,
  checkID string,
  target config.TargetKind,
  metav1.ObjectMeta,
  containerName string,
  isInitContainer bool)
```

so we'll just be passing in names/metadata rather than the underlying objects.

No worries if that's a pain though

rbren (commented):
This is looking awesome! LMK if you want to chat through the frontend changes at all.

rbren (commented):
Looking great! Seems to be a bug when running locally though - I think we need to check `if PodResult != nil` in a few places.

To test locally, you can run:
```
kubectl create ns polaris
helm upgrade --install polaris fairwinds-stable/polaris --set dashboard.ingress.enabled=true -n polaris --set dashboard.ingress.hosts[0]="foo.bar.com"
go run main.go dashboard
```

Probably worth adding a test case to the webhook tests as well

rbren (approved):

PR #470, opened 2021-01-04
Update dependencies

rbren:
This should take care of the dbot PRs. Some breaking changes affected how we do plumbing for the test fixtures, but otherwise should be good.


baderbuddy (approved):
LGTM

PR #471, opened 2021-01-04
Set full object ObjectMeta on new workload from Pod

jordandoig:


baderbuddy:
This looks fantastic to me. Did you test it out and see it working by chance?

rbren (approved):
@baderbuddy @jordandoig any idea how we can test that it works? Is it possible to add a test to our webhook e2e tests?

baderbuddy:
Oh @rbren that's a fantastic idea. I think if we add a test that looks like a failing test but with the annotations to exempt the issues then it should exercise this quite nicely.

PR #483, opened 2021-01-22
Fix a small typo in the README file where a sentence is repeated.

Fairmay:


jordandoig (approved):
Good catch, thanks May!

PR #486, opened 2021-01-29
Added listening address flag

### LINKED ISSUE #440: Make it possible to specify listen address, like 127.0.0.1
**Is your feature request related to a problem? Please describe.**
We have clusters where pods have reachable IP addresses and we currently consider a design where:
* Polaris is deployed as a normal deployment, ideally listening on address 127.0.0.1:8080
* An Oauth proxy is running as a sidecar listening on eg address 0.0.0.0:3000
* All users accessing Polaris is going through the Oauth authenticating proxy to reach Polaris. The Oauth proxy will ensure that the user is authenticated and authorized through group membership

We can protect Polaris even when it's listening on 0.0.0.0:8080 with network policies, but network policies have (interestingly enough) been somewhat unreliable with our provider. It would be very nice if we could make Polaris listen on 127.0.0.1 so that we know for sure that no-one can connect remotely.

**Describe the solution you'd like**
Add an option listenAddress or something similar in addition to the port option. I assume those options would have to be mutually exclusive.

**Describe alternatives you've considered**
Rely on network polices to deny access to the Polaris container port 8080/tcp, use a high port number which is hard to guess or a combination of both.

**Additional context**

makoscafee:
Fixes #440


rbren (approved):
🎉 Thanks Barnie! Think we just need to plumb it through to the helm chart now

PR #487, opened 2021-01-29
Audit- Only show the failed test if flag is specified.

### LINKED ISSUE #433: Outputting only failed tests
Good day! We use polaris v1.2.1 CLI for testing our manifest files in manual mode. The most desirable feature of the application (of course, after analyzing services, ingresses, etc. =)) for us is the ability to display results in the form: score - a list of failed tests. It seems to us that such a conclusion (without a general list of successfully passed tests) greatly simplifies the daily work with the application.

makoscafee:
Fixes #433


pkg/validator/output.go
```diff
@@ -67,15 +75,43 @@ type Result struct {
 	CreatedTime time.Time
 }
 
+func (res *Result) removeSuccessfulResults() {
+	for k, resultMessage := range res.Results {
+		if resultMessage.Success {
```
rbren, line 80: I think we can define this logic once for `ResultSet`, and then call `res.Results.removeSuccessfulResults()` here and below

i.e.
```go
func (rs *ResultSet) {
  for k, resultMessage := range res.Results {
    if resultMessage.Success {
      delete(rs, k)
    }
  }
}
```

rbren:
Thanks Barnie! LGTM - can you just document the flag in `docs/infrastructure-as-code.md`?

rbren (approved):
Thanks for the docs!

PR #492, opened 2021-02-03
add insights

rbren:


baderbuddy (approved):
I think the image scan wouldn't work here unless you do the docker pull first manually. But I'm not positive.

PR #493, opened 2021-02-04
🛠  Add GitHub Action

mambax:
I added a Github Action to add polaris as executable to the GitHub Action runners.

The script downloads the specified polaris version (by tag) and links it to the runners' path.

It is rather verbose and written in a startup-like code manner but it should be well enough for a first version. We already use my action in our CI now (https://github.com/mambax/setup-polaris - a clone before contribution) and it works as planned 🤝 

Some resources:
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-actions
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

Note: In this, I also created https://github.com/github/docs/pull/3468 towards the Github Docs because I nearly went crazy because my Dockerfile_Action did not work (just as a hint for later).

Btw: Once merged to your side and you being happy with it, you can list it in the GH marketplace also.


docs/setup-action.md
```diff
@@ -0,0 +1,37 @@
+# Setup polaris action
```
rbren, line 1: Can we add these to `docs/infrastructure-as-code.md`? That way it gets automatically pulled into the docs at polaris.docs.fairwinds.com

README.md
```diff
@@ -49,6 +49,9 @@ is a platform for auditing Kubernetes clusters and enforcing policy. If you'd li
 
 you can sign up for a [free account here](https://insights.fairwinds.com?source=polaris).
 
+
+## Github Action
+See [our action readme](docs/setup-action.md) for our Github Action.
```
rbren, line 54: We can probably get rid of this with the change below

rbren (approved):
🎉 Thanks for putting this together!

mambax:
@rbren Action is ready. It passes its own test at https://github.com/mambax/polaris/actions/runs/544958628

Note: I also added a second test that only works once its in your repo.

I would also like to request your team or you to once more the action is merged create an issue for yourselves to look at my scripts. I wrote them in a "hurry" (we need polaris finally in our CI 🤣 ) and I am pretty sure they have huge improvement potential. Also I guess my additional Dockerfile could be replaced by one baked out of your build process.

Anyway, let me know any feedback or request more changes.

mambax:
A team member of mine has found that it does not yet work as expected, will update it.

mambax:
It was an internal problem for us when setting up the runners, please review or merge 🍻

rbren:
Thanks @mambax this looks great! Ready to merge I think - mind just signing the CLA?

mambax:
@rbren Emmmmm, how does one do that? ^^

dominikmeyersap:
@rbren MADE IT

rbren:
🎉 Thanks!

PR #495, opened 2021-02-08
enable action

rbren:


baderbuddy (approved):
LGTM

rbren:
@mambax looks like this uses the `Dockerfile` in the root of this repo, rather than the one you created. I think yours is probably the better one - any particular reason for the `build-ext` job rather than just using `build-int`?

.github/workflows/test_setup.yml
```diff
@@ -13,14 +14,14 @@ jobs:
...
-#          version: 3.0.3
-#      - name: Use command
-#        run: polaris version
+
+  build-ext:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup polaris
+        uses: fairwindsops/polaris@master
```
mambax, line 23:
```suggestion
        uses: fairwindsops/polaris/.github/actions/setup-polaris@master
```
I think it has to be like this.
This afaik is the drawback of having the action in the same repo as the "source".

mambax:
> any particular reason for the build-ext job rather than just using build-int

Yes I wanted to make sure it also works for "others" not only internally. I know the reason though I think.

See my suggestion. Afaik if you want it to be so pure as I had it, you need an "action repo". I did not see any other solution yet. See https://github.com/actions/hello-world-docker-action

PR #496, opened 2021-02-11
fix checks for k8s defaults

rbren:
These capabilities are on-by-default, so they need to be dropped, not just not-added.


baderbuddy (approved):
Wow! I knew there were a lot of permissions on by default, but some of those are really surprising

baderbuddy (approved):
LGTM

PR #520, opened 2021-03-22
Add support for check templates

rbren:


pkg/config/schema.go
```diff
@@ -126,38 +148,61 @@ func validateRange(path string, limit interface{}, data interface{}, isMinimum b
...
+		check.SchemaString = string(jsonBytes)
 	}
-	return nil
+	err := json.Unmarshal([]byte(check.SchemaString), &check.Validator)
+	return err
+}
+
+// TemplateForResource fills out a check's templated fields given a particular resource
+func (check SchemaCheck) TemplateForResource(res interface{}) (*SchemaCheck, error) {
+	newCheck := check
```
baderbuddy, line 164: Why create another variable for the check?

baderbuddy (approved):
Looks good to me.

pkg/config/schema.go
rbren: Good question! It's actually making a copy so we can safely modify the schema. Added a comment since that's non-obvious.

PR #524, opened 2021-04-13
Support multi-resource templates

rbren:


pkg/config/schema.go
```diff
@@ -36,35 +37,46 @@ var HandledTargets = []TargetKind{
...
+			return fmt.Errorf("Decoding schema check failed: %v", err)
 		}
 	}
+	return nil
+}
+
+// ParseCheck parses a check from a byte array
+func ParseCheck(id string, rawBytes []byte) (SchemaCheck, error) {
+	check := SchemaCheck{}
+	err := unmarshalYAMLOrJSON(rawBytes, &check)
```
makoscafee, line 76: This will be a bit confusing because it will can accept both `check` and `&check` only to fail during runtime.

pkg/config/schema.go
```diff
@@ -155,27 +167,64 @@ func (check *SchemaCheck) Initialize(id string) error {
...
 	}
 
-	newCheck.SchemaString = w.String()
-	newCheck.Initialize(newCheck.ID)
+	newCheck.AdditionalValidators = map[string]jsonschema.RootSchema{}
+	for kind, schemaStr := range newCheck.AdditionalSchemaStrings {
+		val := jsonschema.RootSchema{}
+		err := unmarshalYAMLOrJSON([]byte(schemaStr), &val)
+		if err != nil {
+			// return nil, err
```
makoscafee, line 219: should be cleaned up now

pkg/validator/summary.go
```diff
@@ -124,10 +124,8 @@ func (c Result) GetSummaryByCategory() CountSummaryByCategory {
 // GetSummary summarizes AuditData
 func (a AuditData) GetSummary() CountSummary {
 	summary := CountSummary{}
-	for _, ctrlResult := range a.Results {
-		if ctrlResult.PodResult != nil {
```
makoscafee, line 128: any reason this check was removed here? i remember we ran into issues in the past without this check.

makoscafee (approved):
LGTM.. just had few comments

pkg/config/schema.go
```diff
@@ -36,35 +37,46 @@ var HandledTargets = []TargetKind{
...
+	Controllers             includeExcludeList                `yaml:"controllers" json:"controllers"`
+	Containers              includeExcludeList                `yaml:"containers" json:"containers"`
+	Target                  TargetKind                        `yaml:"target" json:"target"`
+	SchemaTarget            TargetKind                        `yaml:"schemaTarget" json:"schemaTarget"`
+	Schema                  map[string]interface{}            `yaml:"schema" json:"schema"`
+	SchemaString            string                            `yaml:"jsonSchema" json:"jsonSchema"`
+	Validator               jsonschema.RootSchema             `yaml:"-"`
+	AdditionalSchemas       map[string]map[string]interface{} `yaml:"additionalSchemas" json:"additionalSchemas"`
+	AdditionalSchemaStrings map[string]string                 `yaml:"additionalSchemaStrings" json:"additionalSchemaStrings"`
+	AdditionalValidators    map[string]jsonschema.RootSchema  `yaml:"-"`
```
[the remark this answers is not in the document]
rbren: good catch!

pkg/config/schema.go
rbren: Yeah agreed - I think go generally has this problem. E.g. [json.Unmarshal](https://golang.org/pkg/encoding/json/#Unmarshal) has the same behavior.

I wish there was a way to specify that you want a _pointer_ to an interface, and not just anything.

pkg/config/schema.go
rbren: Oops good catch - removed the panic below as well.

pkg/validator/summary.go
rbren: Yes this is important actually - `res` still has a summary even if it doesn't have a `PodResult`. We're already checking for `PodResult != nil` in the `GetSummary` function so it should be safe.

rbren (commented):
Thanks for the review barnie! Will wait a bit in case anyone else has feedback

checks/missingPodDisruptionBudget.yaml
```diff
@@ -0,0 +1,39 @@
...
+  '$schema': http://json-schema.org/draft-07/schema
+  type: object
+  properties:
+    metadata:
+      type: object
+      properties:
+        labels:
+          type: object
+          minProperties: 1
+additionalSchemaStrings:
```
baderbuddy, line 18: Why is this a string instead of an object?

checks/missingPodDisruptionBudget.yaml
```diff
@@ -0,0 +1,39 @@
...
+  PodDisruptionBudget: |
```
baderbuddy, line 19: Do we specify the API Group?

pkg/config/schema.go
```diff
@@ -155,27 +167,63 @@ func (check *SchemaCheck) Initialize(id string) error {
...
 		check.SchemaString = string(jsonBytes)
 	}
-	err := json.Unmarshal([]byte(check.SchemaString), &check.Validator)
-	return err
+	for kind, schema := range check.AdditionalSchemas {
+		jsonBytes, err := json.Marshal(schema)
+		if err != nil {
+			return err
+		}
+		check.AdditionalSchemaStrings[kind] = string(jsonBytes)
```
baderbuddy, line 175: Since AdditionalSchemas and ADditionalSchemaStrings can both be specified we could be overwriting a schema string here.

pkg/validator/schema.go
```diff
@@ -290,6 +310,20 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
 	if err != nil {
 		return nil, err
 	}
+	for kind := range check.AdditionalValidators {
+		if !passes {
+			break
+		}
+		resources := test.ResourceProvider.Resources[kind]
+		objects := make([]interface{}, len(resources))
```
baderbuddy, line 318: We already have funk here, you could do this with a `Map`

rbren (commented):
Ugh you're totally right on the api version - dunno how I forgot about that.

pkg/config/schema.go
rbren: Documented

pkg/validator/schema.go
rbren: duly funked

rbren:
Now we require Group to be specified as well as Kind. Thanks for pointing that out Bader!

PR #527, opened 2021-04-22
Helper tests2

cowens87:
Add test for all helper methods

### DIFF
--- a/pkg/dashboard/helpers_test.go
+++ b/pkg/dashboard/helpers_test.go
@@ -1,0 +1,35 @@
+package dashboard
+
+import (
+	"testing"
+
+	"github.com/fairwindsops/polaris/pkg/validator"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetGrade(t *testing.T) {
+	input := validator.CountSummary{
+		Successes: 10,
+		Warnings:  3,
+		Dangers:   1,
+	}
+	expectedOutput := "B-"
+	actual := getGrade(input)
+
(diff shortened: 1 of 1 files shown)


pkg/dashboard/helpers_test.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func TestGetIcon(t *testing.T) {
+	input := validator.ResultMessage{
+		ID:       "",
+		Message:  "",
+		Details:  []string(nil),
+		Success:  false,
+		Severity: "",
+		Category: "",
```
rbren, line 142: You can omit all these fields, and just do `input := validator.ResultMessage{}`

In go, every type has a "zero value", i.e. `0` for integers, `0.0` for floats, `""` for strings, `false` for booleans, `[]` for slices, etc. When creating a struct, those values get used if you don't specify anthing.

pkg/dashboard/helpers_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/fairwindsops/polaris/pkg/validator"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWarningWidth(t *testing.T) {
+	input1 := validator.CountSummary{
+		Successes: 0,
+		Warnings:  0,
+		Dangers:   0,
+	}
```
rbren, line 16: Good thought testing the zero-case!

rbren (approved):

PR #529, opened 2021-04-29
Liveness probe

cowens87:
completed success/failure check tests for the following:
- cpuLimitsMissing
- cpuRequestMissing
- livenessProbeMissing
- readinessProbeMissing
- hostPortSet


test/checks/readinessProbeMissing/success.yaml
```diff
@@ -0,0 +1,21 @@
...
+    test: readiness
+  name: readiness-exec
+spec:
+  containers:
+  - name: readiness
+    image: k8s.gcr.io/busybox
+    args:
+    - /bin/sh
+    - -c
+    - touch /tmp/healthy; sleep 30; rm -rf /tmp/healthy; sleep 600
```
baderbuddy, line 14: The readiness probe tests that the pod is ready to go. So maybe instead doing the sleep 30 first then the touch and getting rid of the rm

baderbuddy (commented):
It looks like you have an extra thing checked in called `readinessProb` we'll want to clean that up and then I had a suggestion on the readinessProbeMissing one.

test/checks/readinessProbeMissing/success.yaml
rbren: This actually comes straight from the k8s documentation 😄

test/checks/readinessProb/success.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - /bin/sh
+    - -c
+    - touch /tmp/healthy; sleep 30; rm -rf /tmp/healthy; sleep 600
+    readinessProbe:
+      exec:
+        command:
+        - cat
+        - /tmp/healthy
+      initialDelaySeconds: 5
+      periodSeconds: 5
```
rbren, line 21: I think this got left in by mistake 😄 

I can't figure out why the tests are passing for this one (since the check name is `readinessProb`) - it should fail with `check not found`

rbren: Figured it out! We were only running the `./pkg` tests

https://github.com/FairwindsOps/polaris/pull/532

rbren (approved):

PR #539, opened 2021-05-11
Fix the installation command by helm

sotoiwa:
The installation by helm will fail because the namespace does not exist.
Added the `--create-namespace` flag to create a namespace if not exist.


rbren (approved):
Thanks!

PR #543, opened 2021-05-12
update notReadOnlyRootFilesystem check

cowens87:
Update file check to include good and bad security context


checks/notReadOnlyRootFilesystem.yaml
```diff
@@ -2,15 +2,49 @@ successMessage: Filesystem is read only
...
-  - securityContext
-  properties:
-    securityContext:
-      required:
-      - readOnlyRootFilesystem
+  definitions:
+    goodSecurityContext:
+      type: object
+      anyOf:
+      - required:
```
rbren, line 12: we can remove the `runAsUser`

rbren (approved):
Thanks Cydnee!

PR #546, opened 2021-05-18
Pod level testing

cowens87:


rbren (commented):
Thanks Cydnee - I think we still need to update the checks themselves though - e.g. `checks/privilegeEscalationAllowed.yaml` needs to get the same `goodSecurityContext` and `notBadSecurityContext` treatment that we did for `notReadOnlyRootFilesystem` here: https://github.com/FairwindsOps/polaris/pull/543

Once you update the checks, I think some of these tests will start failing

PR #556, opened 2021-05-27
process helm templates method

cowens87:


cmd/polaris/audit.go
```diff
@@ -77,6 +86,32 @@ var auditCmd = &cobra.Command{
...
+	}
+	params := []string{
+		"template", helmChart,
+		helmChart,
+		"--output-dir",
+		dir,
+	}
+	cmd = exec.Command("helm", params...)
+	err = cmd.Run()
+	if err != nil {
```
rbren, line 109: It looks like the `err` here has some extra info that doesn't get logged automatically when we `panic`. Let's add some more logging to show the `stderr` output when the command fails.

For  background, [here are the docs on cmd.Run()](https://golang.org/pkg/os/exec/#Cmd.Run). You can see there that sometimes the `error` it returns is an [ExitError](https://golang.org/pkg/os/exec/#ExitError)

What we need to do is:
* check if this is an ExitError
* if so, log stderr

tl;dr - I think we want to add this code:
```go
if err != nil {
  if exitErr, ok := err.(exec.ExitErrror); ok {
    logrus.Error(string(exitErr.Stderr))
  }
  return "", err
}
```

to test it out, you can try passing in `--helm-chart=/a/path/that/doesnt/exist`

cmd/polaris/audit.go
```diff
@@ -77,6 +86,32 @@ var auditCmd = &cobra.Command{
...
+	dir, err := ioutil.TempDir("", "*")
+	if err != nil {
+		return "", err
+	}
+	params := []string{
+		"template", helmChart,
+		helmChart,
+		"--output-dir",
+		dir,
+	}
```
rbren, line 106: Can we also add an optional flag `--helm-values`? Then we can add here
```go
if helmValues != "" {
  params = append(params, "--values", helmValues)
}
```

rbren (approved):

PR #563, opened 2021-06-21
fix score when onlyOutputFailed is true

rbren:


cmd/polaris/audit.go
```diff
@@ -76,7 +75,19 @@ var auditCmd = &cobra.Command{
...
+			os.Exit(1)
+		}
+
+		auditData, err := validator.RunAudit(config, k)
+		if err != nil {
+			logrus.Errorf("Error while running audit on resources: %v", err)
+			os.Exit(1)
+		}
+
+		outputAudit(auditData, auditOutputFile, auditOutputURL, auditOutputFormat, useColor, onlyShowFailedTests)
```
makoscafee, line 90: `outputAudit` returns `validator.AuditData` but i do not see it assigned to anything here.

makoscafee (approved):
LGTM... just had one question.

cmd/polaris/audit.go
rbren: Good catch! I wish this was a compile/lint error

PR #586, opened 2021-07-13
Fixed repeated names on dashboard

jdesouza:


rbren (approved):
Good find James! That was a tricky bug.

PR #593, opened 2021-07-23
better support for namespaces in additional schemas

rbren:
This will let you create a check on a `Namespace` object to check for the presence of a resource in that namespace

Discussion here: https://fairwindscommunity.slack.com/archives/CURU6LA91/p1627066123055100


makoscafee (approved):
This LGTM
quick question.. nothing needs to be added in `check.go`?

rbren:
I removed the new check for alertmanager. Since it's a CRD, I think it will break clusters that don't have it.

PR #595, opened 2021-07-27
bump version, add docs, fix up dashboard

rbren:


makoscafee (approved):
This looks good... but the namespace check does not mention the namespaces that do not have that resources.

rbren:
Sorry Barnie, not sure I understand - can you elaborate?

makoscafee:
> Sorry Barnie, not sure I understand - can you elaborate?

So say you are looking for monitoring configuration in each namespace ... meaning if the namespace does not have monitoring configuration and is not in exempted namespace list the check should fail. So say two namespaces do not have monitoring configuration ... in the results you will see two failure but  in the details of each failure you wont see exact namespace missing configuration.

rbren:
Here's what it looks like @makoscafee 
Screen Shot 2021-08-18 at 2 52 04 PM

Issue #635, opened 2021-09-21
Define exemptions configuration precedence


renaudhager:
**Describe the solution you'd like**
Hi guys,

I'm testing Polaris and I'm really enjoying it. 

I made some tests, and unless I made a mistake, it looks like that check exemptions defined in annotations override the exemptions defined in the configuration file.

From security perspective I think it's not ideal, since people that can deploy stuff in a namespace can change the annotation of deployment to exempt a particular check that is not exempt in the configmap, where the access to configmap can be restricted to some specific user it's runs in a dedicated namespace.

I think it it would be nice if the exempt defined in the config file have precedence over the one defined in annotations.

**Describe alternatives you've considered**
Another solution (maybe easier to implement) could to have a flag that completely disable exemptions from annotations.

rbren:
Thanks for the PR!

In terms of precedence, exemptions are the _union_ of annotation and config exemptions. So neither has precedence over the other.

I think your solution of allowing users to disable one or the other makes a lot of sense. Thanks again!

PR #636, opened 2021-09-22
config: new flags '--disallow-(config|annotation)-exemptions'

mvisonneau:
This change follows up #635 and lets end-users decide to disallow exemption rules defined as part of the config file or the controller annotations (whether none, any or both). The main use case here is to be able to prevent users with edit privileges over a controller to add a new exemption rule through an annotation which may obfuscate the actual policies we want to enforce.

### DIFF
--- a/cmd/polaris/root.go
+++ b/cmd/polaris/root.go
@@ -25,7 +25,7 @@ import (
 )
 
 var configPath string
-var disallowExemptions bool
+var disallowExemptions, disallowConfigExemptions, disallowAnnotationExemptions bool
 var logLevel string
 var auditPath string
 var displayName string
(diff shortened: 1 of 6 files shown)


rbren (commented):
Thanks for the contribution! Agreed this is a really useful change.

I do think we need to keep the `--disallow-exemptions` flag for backward compatibility though.

For consistency, I'd also suggest changing the flag names to `--disallow-config-exemptions` and `--disallow-annotation-exemptions`

Otherwise this looks great!

mvisonneau:
:wave: hey @rbren, thanks for the feedback! I proceeded to the changes you recommended, let me know if there is anything else I may have missed :bow:

rbren (approved):
🎉 This looks great! Thanks a bunch

Issue #659, opened 2021-10-29
Incorrect URL to download Polaris tar


Jaanki:
### Installation Process
Creating a Polaris GH action
https://github.com/Jaanki/submariner-operator/pull/3

### Polaris Version
4.2.0

### Expected Behavior
Action to be successfully created

### Actual Behavior
```
/get_polaris.sh: line 6: polaris: command not found
Installing polaris to path.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100     9  100     9    0     0     51      0 --:--:-- --:--:-- --:--:--    51
gzip: invalid magic
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

PR #660, opened 2021-10-29
Correct URL to download Polaris tar

Jaanki:
Closes: https://github.com/FairwindsOps/polaris/issues/659

Signed-Off-By: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/.github/workflows/test_setup.yml
+++ b/.github/workflows/test_setup.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Setup polaris
         uses: ./.github/actions/setup-polaris
         with:
-          version: 3.0.3
+          version: 4.2.0
       - name: Use command
         run: polaris version
 
@@ -22,6 +22,6 @@ jobs:
       - name: Setup polaris
         uses: fairwindsops/polaris/.github/actions/setup-polaris@master
         with:
-          version: 3.0.3
+          version: 4.2.0
       - name: Use command
         run: polaris version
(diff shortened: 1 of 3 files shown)


Jaanki:
See https://github.com/Jaanki/submariner-operator/pull/3 for a successful action creation after making these changes.

rbren (approved):

sudermanjr (changes requested):
It looks to me like we need to update the version in the actions file to the latest version as well.

The reason the `build-ext` job works is that it's pulling the old version in the file (3.0.3) rather than the latest version (which has the new file naming)

So as part of this PR, can you please update these two lines to say 4.2.0:
https://github.com/FairwindsOps/polaris/blob/e31f3f1b4156233cee0abfc38bd7118406c4e181/.github/workflows/test_setup.yml#L14

https://github.com/FairwindsOps/polaris/blob/e31f3f1b4156233cee0abfc38bd7118406c4e181/.github/workflows/test_setup.yml#L25

.github/workflows/test_setup.yml
```diff
@@ -22,6 +22,6 @@ jobs:
       - name: Setup polaris
         uses: fairwindsops/polaris/.github/actions/setup-polaris@master
```
Jaanki, line 23: I think build-ext job will fail until this PR is merged because the job checks out master branch which does not have these changes.

sudermanjr (approved):

PR #678, opened 2021-11-23
skip incomplete or broken YAML - warn user

### LINKED ISSUE #667: Polaris fails to parse/ignore Kustomization yaml JSONPatch files
### Installation Process
Docker image

### Polaris Versions
1.2, 2.0, 3.2.1

### Expected Behavior

It should have ignored the json patch format yaml files when it realized they weren't in the expected yaml format, or there should be an option to ignore files that match a given glob or regex.

### Actual Behavior

Polaris errored out and there doesn't appear to be any CLI way to get it to ignore these jsonpatch yaml files which are valid in Kustomize.

### Steps to Reproduce

```
docker run -ti \
  -v "$PWD:/pwd" \
  -v ~/.kube/config:/opt/app/config:ro \
  quay.io/fairwinds/polaris:1.2 polaris audit \
    --kubeconfig /opt/app/config \
    --audit-path /pwd 
ERRO[0000] Invalid YAML: - op: replace
  path: /spec/syncPolicy
  value: {} 
ERRO[0000] Error parsing YAML: (error unmarshaling JSON: while decoding JSON: json: cannot unmarshal array into Go value of type kube.k8sResource) 
ERRO[0000] Error fetching Kubernetes resources error unmarshaling JSON: while decoding JSON: json: cannot unmarshal array into Go value of type kube.k8sResource 

```

intrand:
This PR:

- changes `visitFile()` (WalkFunc for `filepath.Walk()`) to log parsing failures and continue (eg, patch files; fixes #667)
- removes error logging of failed parsing - upper functions are now responsible for bubbling up the parse error and reporting it
- corrects test for parsing resources to work again :)


pkg/kube/resources_test.go
```diff
@@ -60,7 +60,7 @@ func TestGetMultipleResourceFromSingleFile(t *testing.T) {
 
 func TestGetMultipleResourceFromBadFile(t *testing.T) {
 	_, err := CreateResourceProviderFromPath("./test_files/test_3")
-	assert.NotEqual(t, nil, err, "CreateResource From Path should fail with bad yaml")
+	assert.Equal(t, nil, err, "CreateResource From Path should fail with bad yaml")
```
rbren, line 63:
```suggestion
	assert.Equal(t, nil, err, "CreateResource From Path should not fail with bad yaml")
```

rbren (approved):

TheMatrix97:
Hey! Can we merge this PR? I think it's a pretty interesing feature

pkg/kube/resources.go
```diff
@@ -426,11 +430,7 @@ func (resources *ResourceProvider) addResourcesFromYaml(contents string) error {
 		if strings.TrimSpace(spec) == "" {
 			continue
 		}
-		err := resources.addResourceFromString(spec)
-		if err != nil {
-			logrus.Errorf("Error parsing YAML: (%v)", err)
-			return err
-		}
+		return resources.addResourceFromString(spec)
```
makoscafee, line 433: this will only add Resource from the first spec and ignore others.

intrand:
signed. looks like there's still a test that's failing @makoscafee

PR #712, opened 2022-02-24
Add mutation field to `imagePolicyNotAlways` 

makoscafee:
This PR fixes #

## Checklist
* [x] I have signed the CLA

The goal of the PR is to add a mutation field for one check and try to update the `YAML` based on the mutation definition.

- [x] Add a mutation  field for pullPolicyNotAlways
- [x] Add a test to the [checks tests](https://github.com/FairwindsOps/polaris/tree/master/test/checks) to verify that the mutation works
  - [x] Apply the mutation to failure.yaml
  - [x] Make sure the result matches success.yaml


checks/pullPolicyNotAlways.yaml
```diff
@@ -9,3 +9,7 @@ schema:
   properties:
     imagePullPolicy:
       const: Always
+mutations:
+  - op: add
+    path: /spec/containers/0/imagePullPolicy
```
rbren, line 14:
```suggestion
    path: /imagePullPolicy
```

rbren: Per offline discussion, let's:

* scope this JSON patch down to the container level to match `target` in the check
* return this JSON Patch inside the `ResultMessage` struct
* once all schemas have been applied and you get back a `Result` struct for each resource, which includes each of the applicable JSON Patches:
  * iterate through the `Result` and apply the patches that should be applied based on the user's configuration
  * you'll need to change the JSON Patch `path` field if it's a `PodResult` or `ContainerResult`
    * e.g. `path` for PodResult is different for CronJobs vs Deployments
    * and `path` for ContainerResult will need to use `ContainerResult.Name` to find the right index in the `containers` array

pkg/validator/schema.go
```diff
@@ -251,6 +253,10 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
 	} else if check.Target == config.TargetPod {
 		passes, issues, err = check.CheckPod(test.Resource.PodSpec)
 	} else if check.Target == config.TargetContainer {
+		containerIndex := funk.IndexOf(test.Resource.PodSpec.Containers, func(value corev1.Container) bool {
+			return value.Name == test.Container.Name
+		})
+		prefix = "/spec/template/spec/containers/" + strconv.Itoa(containerIndex)
```
rbren, line 259: I think we need to make this dynamic based on the Kind. E.g. this works for deployments, but not CronJobs

makoscafee: there is an interesting error for that path -
```
add operation does not apply: doc is missing path: "/spec/template/spec/containers/0/imagePullPolicy"
```
unless you use `spec/containers/` need to test for cronjobs to see how it works

examples/config.yaml
```diff
@@ -28,6 +28,9 @@ checks:
   hostPortSet: warning
   tlsSettingsMissing: warning
 
+mutations:
+  - pullPolicyNotAlways
```
rbren, line 32: let's remove from the default config for now, so it's not enabled for everyone yet

pkg/validator/schema.go
```diff
@@ -251,6 +253,18 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
...
+		})
+		if test.Resource.Kind == "CronJob" {
+			prefix = "/spec/jobTemplate/spec/template/spec/containers/" + strconv.Itoa(containerIndex)
+		} else if test.Resource.Kind == "Pod" {
+			// Pod
+			prefix = "/spec/containers/" + strconv.Itoa(containerIndex)
+		} else {
+			// Job, Deployment, Daemonset, ReplicationController and Statefulset
+			prefix = "/spec/template/spec/containers/" + strconv.Itoa(containerIndex)
+		}
```
rbren, line 267: we need similar logic above for `check.Target == TargetPod`

Looks like the if/else logic is kind of complicated. We might want to start a new conditional that says
```go
if check.Target == TargetContainer {
  // logic you have above
} else if check.Target == TargetPod {
  // similar logic, but removing /containers/* from the end
}
```

pkg/validator/schema.go
```diff
@@ -251,6 +253,18 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
...
 	} else if check.Target == config.TargetContainer {
+		containerIndex := funk.IndexOf(test.Resource.PodSpec.Containers, func(value corev1.Container) bool {
+			return value.Name == test.Container.Name
+		})
+		if test.Resource.Kind == "CronJob" {
+			prefix = "/spec/jobTemplate/spec/template/spec/containers/" + strconv.Itoa(containerIndex)
+		} else if test.Resource.Kind == "Pod" {
+			// Pod
+			prefix = "/spec/containers/" + strconv.Itoa(containerIndex)
+		} else {
```
rbren, line 264: I think this should explicitly say `else if kind == Deployment || kind == DaemonSet ...`

The Kind here could hypothetically be any CRD. Maybe we need to just `log.Warning` in the `else`

rbren:
Looking good Barnie! Mostly nits

pkg/validator/schema.go
```diff
@@ -250,7 +253,30 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
 		}
 	} else if check.Target == config.TargetPod {
 		passes, issues, err = check.CheckPod(test.Resource.PodSpec)
+		if test.Resource.Kind == "CronJob" {
```
rbren, line 256: Looking much better!

Can we just extract out a `getJSONSchemaPrefix` function to clean it up a little?

rbren (approved):
Thanks Barnie! one suggestion

PR #718, opened 2022-03-14
bounce out once we hit a type we know to avoid CRD problems

nobletrout:
This PR fixes #

## Checklist
* [X] I have signed the CLA

CRD's cause issues with finding out the pod security contexts of them (since they might be wrapped in a different format than "containers" array

After discussion on slack, added an exclusion list for quick exiting at final state for deployments, daemonsets, cronjobs, etc...

ref: https://fairwindscommunity.slack.com/archives/CURU6LA91/p1647271688336659

Parse and dissect the meta data from a release of a helm chart, or work backwards from parents back to children once no resources are discovered on the parent chart.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Binaries for programs and plugins
+.go-version
 *.exe
 *.exe~
 *.dll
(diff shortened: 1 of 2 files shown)


sudermanjr:
@nobletrout looks like this is your first contribution to Polaris! Thanks for contributing. I'll make sure to bring this to the codeowner's attention

PR #727, opened 2022-03-17
merge the list of resources from custom checks and the generated cont…

staerion:
…roller list before deduplicating them

This PR fixes #
Controllers being shown twice in the Polaris dashboard:

1. Currently the logic used by the Polaris dashboard collects all existing pods of the live cluster. It will then keep check for the parent controller by checking the ownerrefences. Once the ultimate parent controller has been found that does not have any ownerreference it will be added to a list. Every pod will return a "ultimate" parent controller. Because multiple pods can belong to the same controller, this list of controllers  will be "deduplicated" by an internal function. Finally this list will be added to the final list of the controllers to be shown in the dashboard. 

2. At the same time, Polaris dashboard allows for custom checks to be defined. We are using this functionality to perform custom checks on kubernetes custom resource definitions like elasticsearch. All these resources (which can also be controllers such as for example elasticsearch) will then be added to the final list of the controllers to be shown in the dashboard.

The controllers are only being deduplicated in case 1. This means that in the case of pods being created by operators that are watching crds we are ending up with duplicate controllers in the dashboard. First, polaris will find that the ultimate controller of the elasticsearch pods is the elasticsearch crd. Then, we also use custom checks to validate the contents of these elasticsearch crds, they will be added again. In this PR I propose to first merge the list of kubernetes resources from case 1 and case 2 and deduplicate the combined list.

## Checklist
* [X ] I have signed the CLA
* [X ] I have updated/added any relevant documentation

Ensure that custom checks defined in Polaris against crds that create pods do not result in Polaris dashboard showing controllers twice.
1. Define a new slice "kubernetesResources". This slice will hold both the dynamically retrieved controllers and the additional kubernetes resources (additionalKinds) that the custom checks are defined against.
2. The additionalKinds are being added to this slice first instead of directly added to the provider.
3. The LoadControllers() function will return the slice of controllers without calling the deduplication function. 
4. The controllers and additionalKinds resources are merged and deduplicated together and assigned back to kubernetesResources.
5. Add the new deduplicated slice to the provider.

Possibly rename of rethink some of the variables or function names. 
duplicate_controllers

### DIFF
--- a/pkg/kube/resources.go
+++ b/pkg/kube/resources.go
@@ -287,6 +287,7 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
 		}
 	}
 
+	var kubernetesResources []GenericResource
 	for _, kind := range additionalKinds {
 		groupKind := parseGroupKind(maybeTransformKindIntoGroupKind(string(kind)))
 		mapping, err := (restMapper).RESTMapping(groupKind)
@@ -305,7 +306,7 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
 			if err != nil {
 				return nil, err
 			}
-			provider.Resources.addResource(res)
+			kubernetesResources = append(kubernetesResources, res)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


staerion:
@rbren @makoscafee
Any comment on the proposed suggestion or any other ideas what would be a way this duplication can be circumvented in the dashboard? Currently any crd with custom polaris checks will appear twice in the dashboard if that crd is the final owner of running pods.

rbren (approved):
This makes sense to me. Appreciate the fix!

PR #734, opened 2022-03-29
Added more mutations and refactor test to test each mutation separately

makoscafee:
This PR fixes #

## Checklist
* [x] I have signed the CLA

### DIFF
--- a/checks/memoryLimitsMissing.yaml
+++ b/checks/memoryLimitsMissing.yaml
@@ -25,3 +25,7 @@ schema:
               type: string
               not:
                 const: ''
+mutations:
+  - op: add
+    path: /resources/limits
+    value: {memory: "512Mi"}
--- a/checks/memoryLimitsMissing.yaml
+++ b/checks/memoryLimitsMissing.yaml
@@ -28,4 +28,7 @@ schema:
 mutations:
   - op: add
     path: /resources/limits
-    value: {memory: "512Mi"}
+    value: {}
+  - op: add
+    path: /resources/limits/memory
+    value: "512Mi"
--- a/checks/memoryRequestsMissing.yaml
+++ b/checks/memoryRequestsMissing.yaml
@@ -28,4 +28,7 @@ schema:
 mutations:
   - op: add
     path: /resources/requests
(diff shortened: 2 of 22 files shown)


pkg/validator/schema.go
```diff
@@ -264,6 +271,7 @@ func applySchemaCheck(conf *config.Configuration, checkID string, test schemaTes
 		}
 		passes, issues, err = check.CheckContainer(test.Container)
 	} else {
+		prefix = "/spec"
```
rbren, line 274: I think this should be removed and put inside the check YAML

checks/memoryLimitsMissing.yaml
```diff
@@ -25,3 +25,7 @@ schema:
               type: string
               not:
                 const: ''
+mutations:
+  - op: add
+    path: /resources/limits
```
rbren, line 30: shouldn't this be `/resources/limits/memory`?

checks/memoryRequestsMissing.yaml
```diff
@@ -25,3 +25,7 @@ schema:
               type: string
               not:
                 const: ''
+mutations:
+  - op: add
+    path: /resources/requests
```
rbren, line 30: ditto

rbren (commented):
For the two that appear to be broken, do we have a test case set up?

checks/memoryLimitsMissing.yaml
```diff
@@ -25,3 +25,7 @@ schema:
...
+    value: {memory: "512Mi"}
```
rbren, line 31: Oh wait I get it now. Why are these different from CPU?

rbren: @makoscafee any reason these are different? I think this check would modify `limits.cpu` as-is

rbren: 🤔 maybe we need two mutations: one to add `limits`, and then one to add `limits.memory`?

makoscafee: Done

rbren (approved):

PR #742, opened 2022-04-05
Add a --namespace flag to the in-cluster audit

### LINKED ISSUE #741: Filter by namespace in CLI
Related to the improvement added for #123 - it'd be nice to be able to do this from the CLI as well without having to manually parse JSON.  

Essentially I'm looking for something like this: 

```shell
polaris audit --namespace foo
```

And have it generate the report for *JUST* the `foo` namespace (same as what happens when you filter namespaces in the dashboard).

sudermanjr:
This PR fixes #741

## Checklist
* [x] I have signed the CLA

Add a namespace flag to the audit command to allow in-cluster auditing of a single namespace

Added Namespace to the config struct, used it in a few places

Not doing this is still an option - see caveats:

### Considerations / Caveats

- [ ] The output probably still needs to be modifed to indicate that these results are not actually clsuter-wide. How best to do this?
- [ ] We will drop any non-namespaced objects I think, such as nodes. Is that okay?


examples/config-full.yaml
```diff
@@ -88,3 +88,4 @@ customChecks:
           not:
             pattern: ^quay.io
 
+namespce: test-ns
```
sudermanjr, line 91:
```suggestion
namespce: test-ns

```

examples/config.yaml
```diff
@@ -202,3 +202,5 @@ exemptions:
       - kube-hunter
     rules:
       - runAsRootAllowed
+
+namespace: test-ns
```
sudermanjr, line 206:
```suggestion
namespace: test-ns

```

makoscafee (approved):

PR #765, opened 2022-05-20
fix: properly remove emojis in pretty format with no color

### LINKED ISSUE #764: Emojis are not fully stripped in pretty format with no colors
### What happened?

Audit in pretty format with no colors does not properly strip emojis leaving some broken binary symbols:
```shell
$ polaris audit --audit-path /tmp/1.yaml -f pretty --color=false
[...]
Deployment app
    deploymentMissingReplicas            �� Success
        Reliability - Multiple replicas are scheduled
```

No special characters should be present next to the status field.

`polaris audit --audit-path <PATH> -f pretty --color=false`

### Version

6.0.0

### Search

### Code of Conduct

### Additional context

_No response_

weisdd:
This PR fixes #764 

## Checklist
* [x] I have signed the CLA

The goal is to fix emojis removal from pretty outputs in colorless mode. Without the fix, you'd get broken binary symbols, e.g.:
```shell
Deployment app
    deploymentMissingReplicas            �� Success
        Reliability - Multiple replicas are scheduled
```

I haven't seen any double-worded status fields, so instead of slicing, just extracted a word after a whitespace.
```go
status = strings.Fields(status)[1]
```

If there's a need to have such double-worded names, we can replace the code with:
```go
status = strings.Join(strings.Fields(status)[1:], " ")
```
or any other implementation. In Go 1.18, to which the project hasn't migrated yet, there's [strings.Cut()](https://pkg.go.dev/strings#Cut), which I am a fan of. :)

Not really an alternative, but currently I have to play with `sed` to remove those symbols.


pkg/validator/output.go
```diff
@@ -211,7 +212,7 @@ func (res ResultSet) GetPrettyOutput() string {
 			}
 		}
 		if color.NoColor {
-			status = status[2:] // remove emoji
+			status = strings.Fields(status)[1] // remove emoji
```
rbren, line 215: I wonder if this would be safer as
```
strings.Join(strings.Fields(status)[1:], " ")
```

rbren: Probably not a big deal

rbren (approved):

pkg/validator/output.go
weisdd: @rbren Please, feel free to adjust it to whatever serves the project best :)

PR #788, opened 2022-07-07
passing the actual path of Polaris to GITHUB_PATH

michaelniemand:
I don't understand how this could have ever worked before, just passing "polaris" to GITHUB_PATH before, not the actual, absolute path?!

This PR fixes #

## Checklist
* [x] I have signed the CLA

the install script does not work, the polaris executable is not found

passing the actual path to the executable to GITHUB_PATH

-

### DIFF
--- a/.github/actions/setup-polaris/get_polaris.sh
+++ b/.github/actions/setup-polaris/get_polaris.sh
@@ -16,5 +16,5 @@ curl -LJ -o $TARGET_FILE $POLARIS_URL
 mkdir polaris
 tar -xzf $TARGET_FILE -C polaris
 rm $TARGET_FILE
-echo "polaris" >> $GITHUB_PATH
+echo "$(pwd)/polaris/polaris" >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


.github/actions/setup-polaris/get_polaris.sh
```diff
@@ -16,5 +16,5 @@ curl -LJ -o $TARGET_FILE $POLARIS_URL
 mkdir polaris
 tar -xzf $TARGET_FILE -C polaris
 rm $TARGET_FILE
-echo "polaris" >> $GITHUB_PATH
-echo "::set-output name=version::$INPUT_VERSION"
\ No newline at end of file
+echo "$(pwd)/polaris/polaris" >> $GITHUB_PATH
```
rbren, line 19: 🤔 is this the right path? Should `polaris` be duplicated?

michaelniemand: you are right, my commit points to the binary, not the directory

PR #797, opened 2022-07-11
Add `checks` flag  to fix specific checks

makoscafee:
This PR fixes #

## Checklist
* [x] I have signed the CLA

To add two flags
- `check` flag to only fix specific checks and ignore others with `check=all` fixing all defined mutations
-


cmd/polaris/fix.go
```diff
@@ -73,6 +79,19 @@ var fixCommand = &cobra.Command{
 		}
 		var contentStr string
 		isFirstResource := true
+
+		if len(checksToFix) > 0 {
+			config.Mutations = checksToFix
```
makoscafee, line 84: do we want to verify these checks supplied to the one in configurations?

rbren (approved):

PR #816, opened 2022-07-29
FWI-2476: Add missingNetworkPolicy, automountServiceAccountToken, and linuxHardening checks

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

Add checks for

* missingNetworkPolicy - requires a NetworkPolicy that contains applied egress and ingress rules, and targets a Pod label.
* automountServiceAccountToken - requires that auto mounting of a ServiceAccount token is disabled for a Pod, either at the ServiceAccount or Pod level.
* linuxHardening - requires that one of seLinux options, AppArmor, sec comp profile, or dropping Linux capabilities are used to prevent a container from using unwanted privileges. Some of these can be set at the Pod or Container level.
Added the above checks along with success and failure schema test-cases to validate fields which can be set or overridden at the pod and container spec.

Internal Ticket: [FWI-2476](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-2476)

### DIFF
--- a/checks/linuxHardening.yaml
+++ b/checks/linuxHardening.yaml
@@ -1,0 +1,89 @@
+successMessage: One of AppArmor, Seccomp, SELinux, or dropping Linux Capabilities are used to restrict containers using unwanted privileges
+FailureMessage: Use one of AppArmor, Seccomp, SELinux, or dropping Linux Capabilities to restrict containers using unwanted privileges
+category: Security
+target: Container
+schemaString: |
+  '$schema': http://json-schema.org/draft-07/schema
+  definitions:
+    podOrContainerSeccompProfile:
+      type: object
+      {{ $podSeccompProfileType := .Polaris.PodSpec.securityContext.seccompProfile.type }}
+      required:
+      {{ if or (not $podSeccompProfileType) (eq $podSeccompProfileType "Unconfined") }}
+        - securityContext
+      {{ end }}
+      properties:
+        securityContext:
+          type: object
+          required:
+          {{ if or (not $podSeccompProfileType) (eq $podSeccompProfileType "Unconfined") }}
+            - seccompProfile
+          {{ end }}
+          properties:
+            seccompProfile:
+              type: object
+              required:
+              {{ if or (not $podSeccompProfileType) (eq $podSeccompProfileType "Unconfined") }}
+                - type
+              {{ end }}
+              properties:
+                type:
+                  type: string
+                  allOf:
+                    - not:
+                        const: "Unconfined"
+                    {{ if or (not $podSeccompProfileType) (eq $podSeccompProfileType "Unconfined") }}
+                    - minLength: 1
+                    {{ end }}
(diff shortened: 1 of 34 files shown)


makoscafee (approved):

checks/linuxHardening.yaml
```diff
@@ -0,0 +1,89 @@
+successMessage: One of AppArmor, Seccomp, SELinux, or dropping Linux Capabilities are used to restrict containers using unwanted privileges
+FailureMessage: Use one of AppArmor, Seccomp, SELinux, or dropping Linux Capabilities to restrict containers using unwanted privileges
+category: Security
+target: Container
```
makoscafee, line 4: Here the target is `Container` but you are checking `SecompProfile` from PodSpec.SecurityContext and not checking for `Containers.SecurityContext.` was this on purpose?

checks/missingNetworkPolicy.yaml
```diff
@@ -0,0 +1,49 @@
...
+        labels:
+          type: object
+          minProperties: 1
+additionalSchemaStrings:
+  networking.k8s.io/NetworkPolicy: |
+    type: object
+    properties:
+      spec:
+        type: object
+        required: ["podSelector", "egress", "ingress"]
```
makoscafee, line 21: egress and ingress are optional any one of these could be present. do we want to enforce the presence of both?

makoscafee (approved):
LGTM, just had few questions .

checks/linuxHardening.yaml
```diff
@@ -0,0 +1,89 @@
...
+          {{ end }}
+          properties:
+            seccompProfile:
+              type: object
+              required:
+              {{ if or (not $podSeccompProfileType) (eq $podSeccompProfileType "Unconfined") }}
+                - type
+              {{ end }}
+              properties:
+                type:
```
ivanfetch-fw, line 30: @makoscafee RE: your comment
> Here the target is Container but you are checking SecompProfile from PodSpec.SecurityContext and not checking for Containers.SecurityContext. was this on purpose?

Here is where the per-container schema requires `seccompProfile.type` depending on what the pod-level field is set to; if the pod-level field is empty or set to `Unconfined` the container field must be set to something other than `Unconfined`.

checks/missingNetworkPolicy.yaml
ivanfetch-fw: In this case yes - the ticket associated with this check requires both.

checks/linuxHardening.yaml
ivanfetch-fw: The per-container schema is requiring `securityContext` depending on the pod-level field. I commented with more detail in line 30 of this file.

ivanfetch-fw:
Barnie [confirmed he doesn't have additional comments and this can be merged](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1659714043413909?thread_ts=1658858841.999019&cid=CNQBGRGFN).

PR #820, opened 2022-08-09
FWI-2547: Add checks for RBAC allowing execing or attaching to a Pod

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

* Alert on user-defined Roles and ClusterRoles that allow execing or attaching to pods.
* Alert on user-defined RoleBindings and ClusterRoleBindings that reference Roles or ClusterRoles allowing execing or attaching to pods.

The ClusterRole check factors out static names of built-in ClusterRoles that allow exec or attach. I chose not to match partial ClusterRole names like `system:*` or standard labels, because it is possible for user-defined ClusterRoles to be created with these same criteria. If future versions of Kubernetes add ClusterRoles with these permissions, those new ClusterRoles will need to be added to this check.

### DIFF
--- a/checks/rolebindingClusterRolePodExecAttach.yaml
+++ b/checks/rolebindingClusterRolePodExecAttach.yaml
@@ -6,7 +6,7 @@ schemaString: |
   '$schema': http://json-schema.org/draft-07/schema
   type: object
   anyOf:
-    # Do not alert on default Roles.
+    # Do not alert on default RoleBindings.
     - required: ["metadata"]
       properties:
         metadata:
@@ -19,6 +19,15 @@ schemaString: |
                 - const: "cluster-admin"
                 - const: "system:controller:generic-garbage-collector"
                 - const: "system:controller:namespace-controller"
+    # Pass RoleBindings that point to a Role.
+    - required: ["roleRef"]
+      properties:
+        roleRef:
+          required: ["kind"]
+          properties:
+            kind:
+              type: string
+              const: "Role"
     - required: ["roleRef"]
       properties:
         roleRef:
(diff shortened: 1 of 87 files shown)


ivanfetch-fw:
I've added a new schema test which will fail until [Polaris PR 822](https://github.com/FairwindsOps/polaris/pull/822) is merged - fixing an issue with namespace filtering Kube resources for additional schema validation.

checks/rolebindingClusterRolePodExecAttach.yaml
```diff
@@ -0,0 +1,94 @@
+successMessage: The RoleBinding does not reference a ClusterRole allowing pods/exec or pods/attach
```
makoscafee, line 1: messaging still refers to `ClusterRole` in role binding

checks/rolebindingClusterRolePodExecAttach.yaml
```diff
@@ -0,0 +1,94 @@
...
+failureMessage: The RoleBinding references a ClusterRole that allows Pods/exec, allows pods/attach, or that does not exist
```
makoscafee, line 2: ditto

makoscafee (approved):
LGTM
just wanted to ask if the test for ClusterRole/ClusterRoleBinding/Role/RoleBinding which has these constants in their names was not included on purpose
```
-"cluster-admin"
- "system:controller:generic-garbage-collector"
- "system:controller:namespace-controller"
etc.
```

checks/rolebindingClusterRolePodExecAttach.yaml
ivanfetch-fw: I believe this message is correct, because the primary schema looks at bindings, and the additional schema looks at the ClusterRole associated with the binding.

checks/rolebindingClusterRolePodExecAttach.yaml
ivanfetch-fw: I believe this message is correct, because the primary schema looks at bindings, and the additional schema looks at the ClusterRole associated with the binding.

ivanfetch-fw:
@makoscafee  Thank you for the review. Let me know whether you have additional things you think I should address, or whether this is good to merge.


> LGTM just wanted to ask if the test for ClusterRole/ClusterRoleBinding/Role/RoleBinding which has these constants in their names was not included on purpose
> -"cluster-admin"
> - "system:controller:generic-garbage-collector"
> - "system:controller:namespace-controller"
> etc.
> ```

I did add a couple more tests for those, thanks for that suggestion.

I also removed code that checks for default bindings to ClusterRoles and Roles, because there actually are not any default bindings in Kube - that code being in those 2 checks was an artifact of copy/pasting from the ClusterRoleBinding check.

makoscafee (approved):

PR #821, opened 2022-08-10
[FWI-2357] Let Polaris modify YAML without losing comments/formatting

makoscafee:
This PR fixes #

## Checklist
* [ ] I have signed the CLA


checks/deploymentMissingReplicas.yaml
```diff
@@ -21,5 +21,5 @@ schema:
           minimum: 2
 mutations:
   - op: add
-    path: /spec/replicas
+    path: spec/replicas
```
rbren, line 24: Why did we lose the `/` here?

checks/hostPortSet.yaml
```diff
@@ -12,4 +12,7 @@ schema:
       items:
         properties:
           hostPort:
-            const: 0
\ No newline at end of file
+            const: 0
+mutations:
+  - op: remove
+    path: /ports[*]/hostPort
```
rbren, line 18: is it possible to make this syntax `/ports/*/hostPort`? That way it is closer to JSON patch.

We could even do something like `strings.Replace("/*/", "[*]/")`

test/checks/cpuRequestsMissing/mutated.yaml
```diff
@@ -1,14 +1,14 @@
...
 spec:
   containers:
-  - image: nginx
-    name: nginx
-    resources:
-      requests:
-        cpu: 100m #TODO: Set this to the amount of CPU you want to reserve for your workload
-        memory: 128Mi
+    - name: nginx
+      image: nginx
```
rbren, line 10: Great to see key ordering preserved!

pkg/mutation/mutate.go
```diff
@@ -2,44 +2,60 @@ package mutation
...
+	yaml "gopkg.in/yaml.v3"
 )
 
-// ApplyAllSchemaMutations applies available mutation to a single resource
-func ApplyAllSchemaMutations(conf *config.Configuration, resourceProvider *kube.ResourceProvider, resource kube.GenericResource, mutations []jsonpatch.Operation) (kube.GenericResource, error) {
-	resByte := resource.OriginalObjectJSON
-	var jsonByte []byte
-	mutationByte, err := json.Marshal(mutations)
+// ApplyAllMutations applies available mutation to a single resource
+func ApplyAllMutations(manifest string, mutations []jsonpatch.Operation) (string, error) {
```
rbren, line 19: It would be great to see some generic tests of this, in addition to the kubernetes ones. That way we can test really complex mutations, like indexing into multiple arrays.

E.g.
```yaml
pets:
- name: fido
  owners:
  - name: Alice
  - name: Bob
     aliases:
     - Robert
```

```
mutations:
- op: add
  path: /pets/0/owners/*/aliases
  value: Rob
```

pkg/mutation/mutate.go
```diff
@@ -116,3 +132,225 @@ func UpdateMutatedContentWithComments(yamlContent string, comments []config.Muta
 	}
 	return fileContent
 }
+
+func findNodes(node *yaml.Node, selectors []string) ([]*yaml.Node, error) {
```
rbren, line 136: This function looks great!

I wonder if we can break it up a bit to make it a little more readable...some comments below

pkg/mutation/mutate.go
```diff
@@ -116,3 +132,225 @@ func UpdateMutatedContentWithComments(yamlContent string, comments []config.Muta
...
+	var nodes []*yaml.Node
+	currentSelector := selectors[0]
+	// array[N] or array[*] selectors.
+	if i := strings.LastIndex(currentSelector, "["); i > 0 && strings.HasSuffix(currentSelector, "]") {
```
rbren, line 140: Maybe put the inside of this `if` in a `findArrayNodes` function?

pkg/mutation/mutate.go
```diff
@@ -116,3 +132,225 @@ func UpdateMutatedContentWithComments(yamlContent string, comments []config.Muta
...
+		node.Tag = "!!map"
+		node.Value = ""
+	case yaml.SequenceNode:
+		return nil, errors.Errorf("parent node is array, use [*] or [0]..[%v] instead of .%v to access its item(s) first", len(node.Content)-1, currentSelector)
+
+	default:
+		return nil, errors.Errorf("parent node is of unknown kind %v", node.Kind)
+	}
+	// Create the rest of the selector path.
+	for _, selector := range selectors {
```
rbren, line 218: It's interesting that `findNodes` _also_ modifies the nodes.

I wonder if we should break out an `ensurePathExists`, or at least rename `findNodes` to something like `createPathAndFindNodes`

pkg/mutation/mutate.go
```diff
@@ -116,3 +132,225 @@ func UpdateMutatedContentWithComments(yamlContent string, comments []config.Muta
...
+			if index >= 0 { // array[N]
+				visitArrayNodes = []*yaml.Node{arrayNode.Content[index]}
+			} else { // array[*]
+				visitArrayNodes = arrayNode.Content
+			}
+			for _, node := range visitArrayNodes {
+				lastSelector := len(selectors) == 1
+				if !lastSelector {
+					removeMatchingNode(node, selectors[1:])
+				}
```
rbren, line 334: I'm not sure I understand this part

pkg/mutation/mutate.go
```diff
@@ -116,3 +132,225 @@ func UpdateMutatedContentWithComments(yamlContent string, comments []config.Muta
...
+
+	err := removeMatchingNode(doc.Content[0], selectors)
+	if err != nil {
+		return errors.Wrapf(err, "failed to match %q", path)
+	}
+
+	return nil
+}
+
+func removeMatchingNode(node *yaml.Node, selectors []string) error {
```
rbren, line 294: Seems like we can probably dedupe some of this logic with `findNodes`

pkg/mutation/mutate_test.go
```diff
@@ -0,0 +1,82 @@
...
+    owners:
+      - name: Alice
+        aliases:
+          - rob
+      - name: Bob
+        aliases:
+          - rob
+          - Robert
+  - name: scooby
+`
```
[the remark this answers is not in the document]
makoscafee: done

rbren (approved):

PR #845, opened 2022-09-19
Build docker images using goreleaser and support arm64

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

* Use goreleaser to build/push Docker images
* Support arm64 docker images
* Only release to Github and push `latest` and version-related Docker tags when the git branch is `master` (requires a goreleaser wrapper script).
* End up updating Polaris to Go 1.19.
*


scripts/goreleaser.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env sh
+# Wrap goreleaser by using envsubst on .goreleaser.yml,
```
rbren, line 2: can you copy the latest from insights-plugins? I made some formatting changes.

Also--do we need this script here? Polaris tagging strategy should be much more normal than the monorepo

ivanfetch-fw: @rbren We do need this script to push docker manifests for feature branches. How does the indenting look now - this script differs from that of insights-plugins, so I fixed indenting by hand.

rbren (approved):

rbren (approved):

PR #866, opened 2022-10-17
Fix conflict between dangerousCapabilities and insecureCapabilities

rbren:
This PR fixes an issue where the mutation for `dangerousCapabilities` interferes with the mutation for `insecureCapabilities`

## Checklist
* [x] I have signed the CLA

Make mutation for dangerousCapabilities more specific

There's a larger issue where mutations run in a non-deterministic order. There's also no defense against mutations doing conflicting or even opposite things (e.g. one might set cpu limits, and another remove cpu limits).

### DIFF
--- a/checks/dangerousCapabilities.yaml
+++ b/checks/dangerousCapabilities.yaml
@@ -28,4 +28,4 @@ schema:
 
 mutations:
   - op: remove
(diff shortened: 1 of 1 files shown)


mggude (approved):
LGTM!

PR #871, opened 2022-11-02
Set the `automountServiceAccountToken` to ignore, `hasPrefix` and `hasSuffix` functions are available in the go template, exempt the prefix `system:` instead of indifidual entries for RBAC checks

ivanfetch-fw:
Note that this PR will merge into another feature branch, which is currently used with a subset of `Insights telemetry` users.

Internal Ticket: https://fairwinds.myjetbrains.com/youtrack/issue/FWI-3096/Update-RBAC-checks-to-exempt-the-system-prefix-instead-of-individual-resource-names

## Checklist
* [x] I have signed the CLA

* Exempt additional resources from RBAC checks that are provided by Kube, or cloud providers.
* Set the `automountServiceAccountToken` to ignore in the default config
* The RBAC checks `clusterrolebindingClusterAdmin` and `clusterrolebindingPodExecAttach` now exempt bindings that start with `system:`, instead of listing explicit strings.
* The RBAC check `clusterrolePodExecAttach` now exempt ClusterRoles that start with `system:`, instead of listing explicit strings.

We could continue to maintain individual exemptions like `system:controller:...`, but this list will likely grow with new Kube releases.

### DIFF
--- a/pkg/config/schema.go
+++ b/pkg/config/schema.go
@@ -223,7 +223,10 @@ func (check SchemaCheck) TemplateForResource(res interface{}) (*SchemaCheck, err
 	newCheck.AdditionalSchemaStrings = map[string]string{}
 
 	for kind, tmplString := range templateStrings {
-		tmpl := template.New(newCheck.ID)
+		tmpl := template.New(newCheck.ID).Funcs(template.FuncMap{
+			"hasPrefix": strings.HasPrefix,
+			"hasSuffix": strings.HasSuffix,
+		})
 		tmpl, err := tmpl.Parse(tmplString)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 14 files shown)


pkg/config/schema.go
```diff
@@ -223,7 +223,10 @@ func (check SchemaCheck) TemplateForResource(res interface{}) (*SchemaCheck, err
 	newCheck.AdditionalSchemaStrings = map[string]string{}
 
 	for kind, tmplString := range templateStrings {
-		tmpl := template.New(newCheck.ID)
+		tmpl := template.New(newCheck.ID).Funcs(template.FuncMap{
+			"hasPrefix": strings.HasPrefix,
+			"hasSuffix": strings.HasSuffix,
+		})
```
rbren, line 229: Huh I expected these to be built-in.

rbren (approved):

PR #879, opened 2022-11-16
Add a check for topologySpreadConstraint

### LINKED ISSUE #547: New checks for anti-affinity policy or topology spread
Polaris needs some additional validational checks like Explicit anti-affinity policy  for production deployments.  
Why?
By default, the Kubernetes scheduler uses a bin-packing algorithm to fit as many pods as possible into a cluster. The scheduler prefers a more evenly distributed general node load to app replicas precisely spread across nodes. Therefore, by default, multi-replica is not guaranteed multi-AZ.

Reason to add: Need a way to check if our prod deployments contains soft anti affinity to enable multi az deployments, something to verify that there is something similar to : 
`    spec:
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - podAffinityTerm:
              labelSelector:
                matchExpressions:
                - key: app
                  operator: In
                  values:
                  - %appnamehere%
              topologyKey: failure-domain.beta.kubernetes.io/zone
            weight: 100`

sudermanjr:
This PR fixes #547

## Checklist
* [x] I have signed the CLA

See issue #547 - Add a check for pod topologySpreadConstraints - recommending that users set these to ensure high availability across zones and/or hosts

Add topologySpreadConstraint check

There may be more parts of the spec that we want to recommend, and possibly only limit to one topologyKey.

Also, possibly splitting this into two checks - one that ensures a topologySpreadConstraint exists, and another to verify its configuration

### DIFF
--- a/checks/topologySpreadConstraint.yaml
+++ b/checks/topologySpreadConstraint.yaml
@@ -1,0 +1,21 @@
+successMessage: Pod has a valid topology spread constraint
+failureMessage: Pod should be configured with a valid topology spread constraint
+category: Reliability
+target: PodSpec
+schema:
+  '$schema': http://json-schema.org/draft-07/schema
+  type: object
+  required:
+    - topologySpreadConstraints
+  properties:
+    topologySpreadConstraints:
+      type: array
+      items:
+        type: object
+        properties:
+          topologyKey:
+            anyOf:
+              - type: string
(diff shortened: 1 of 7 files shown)


rbren (approved):
Thanks for adding tests!

PR #909, opened 2023-02-02
Fix #908 by generating release name

### LINKED ISSUE #908: Polaris giving helm chart directory to helm as release name
### What happened?

Polaris audit failed because the helm chart's directory is an invalid release name.

That is, this command led to this error:

```bash
polaris audit --helm-chart helm/charts/proj --helm-values helm/values/test_values.yaml

ERRO[0002] Error: release name "helm/charts/proj": invalid release name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 53 
INFO[0002] Couldn't process helm chart: exit status 1 
```

I expected the chart directory to not be substituted in as the helm release name.

I expect it should be easy enough. Looks like the chart directory is being substituted as the helm release name.

### Version

7.3.0

### Search

### Code of Conduct

### Additional context

_No response_

JoeHCQ1:
This PR fixes #908 

## Checklist
* [x] I have signed the CLA

The goal is to fix the helm chart auditing so it works for what I am wanting to do today (also makes it work per [the docs](https://github.com/FairwindsOps/polaris/blob/master/docs/infrastructure-as-code.md#audit-helm-charts)).

I change the helm template command to have helm generate a release name instead of using the helm chart path as the release name (which is frequently not a valid release name).

You may want to hard-code the release name to something more self-explanatory if it shows up in the output like `polarisReleaseName`.

### Testing

I tested the changes locally, and Polaris audit now works instead of breaking quickly. Note, I also ran `go test`.

### DIFF
--- a/cmd/polaris/audit.go
+++ b/cmd/polaris/audit.go
@@ -143,7 +143,7 @@ func ProcessHelmTemplates(helmChart, helmValues string) (string, error) {
 	}
 	params := []string{
 		"template", helmChart,
-		helmChart,
+		"--generate-name",
 		"--output-dir",
 		dir,
 	}


rbren (approved):
Makes sense! Thanks.

PR #919, opened 2023-02-16
Add kubeconfig path

fengshunli:
This PR fixes #

## Checklist
* [x] I have signed the CLA

add kubeconfig parameter, you can switch clusters

### DIFF
--- a/pkg/kube/resources.go
+++ b/pkg/kube/resources.go
@@ -20,6 +20,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -135,7 +137,7 @@ var podSpecFields = []string{"jobTemplate", "spec", "template"}
 // CreateResourceProvider returns a new ResourceProvider object to interact with k8s resources
 func CreateResourceProvider(ctx context.Context, directory, workload string, c conf.Configuration) (*ResourceProvider, error) {
 	if workload != "" {
-		return CreateResourceProviderFromResource(ctx, workload)
+		return CreateResourceProviderFromResource(ctx, workload, c.KubeConfig)
 	}
 	if directory != "" {
 		return CreateResourceProviderFromPath(directory)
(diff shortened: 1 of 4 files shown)


pkg/kube/resources.go
```diff
@@ -242,12 +255,28 @@ func CreateResourceProviderFromYaml(yamlContent string) *ResourceProvider {
...
+			logrus.Errorf("Error fetching KubeConfig: %v", err)
+			return nil, err
+		}
+	} else {
+		kubeConf, err = config.GetConfig()
+		if err != nil {
+			logrus.Errorf("Error fetching KubeConfig: %v", err)
+			return nil, err
+		}
+	}
```
rbren, line 279: Do you think we can extract all this logic out into a helper function? That way it's not duplicated here and `CreateResourceProviderFromResource`. (Also it looks like we missed the `KubeContext` option there!)

rbren (commented):
Thanks for the contribution! This looks like a very helpful change.

FWIW I _think_ you can get the same functionality today by setting the `KUBECONFIG` env var.

pkg/kube/resources.go
fengshunli: KubeContext is specified by specifying kubeconfig, without kubeconfig, KubeContext is meaningless

rbren:
Hmm--not quite sure what you mean. How would `KUBECONFIG=~/kconf polaris audit` differ from `polaris audit --kubeconfig=~/kconf`?

You might be right, just want to make sure I understand

fengshunli:
Right now I'm using --kubeconfig
image

fengshunli:
> Hmm--not quite sure what you mean. How would `KUBECONFIG=~/kconf polaris audit` differ from `polaris audit --kubeconfig=~/kconf`?
> 
> You might be right, just want to make sure I understand

Do you mean that I need to change --kubeconfig to KUBECONFIG @rbren

rbren:
No, no need to change anything in this PR. I was just saying that running
```
export KUBECONFIG=./kconf
polaris audit
```
or more simply
```
KUBECONFIG=./kconf polaris audit
```
are both supported today. I think that achieves the same thing if you wanted a workaround.

I like having the `--kubeconfig` flag too though!

pkg/kube/resources.go
rbren: Right exactly. But we're not using `KubeContext` in `CreateResourceProviderFromResource`, even though we have a kubeconfig there.

fengshunli:
> KUBECONFIG

KUBECONFIG is also available for the method

PR #936, opened 2023-04-03
Update checks documentation

transient1:
update checks page with newer checks. Make sure config-full.yaml has all of the checks in config.yaml. Update list of checks that can be mutated.


rbren (approved):
:woot: thanks Stevie!

PR #966, opened 2023-06-21
Fix nil pointer issue with webhook

rbren:
This PR hopefully fixes a `panic` in the admission controller

## Checklist
* [x] I have signed the CLA

Some users report a `nil pointer` panic when running the webhook as of 8.2


.circleci/config.yml
```diff
@@ -161,8 +161,6 @@ workflows:
               only: /.*/
       - build_and_push:
           context: org-global
-          requires:
```
jdesouza, line 164: Was that intentional to remove the test?

rbren: Yeah I don't think it's necessary.

rbren: tests still run--they just don't block building the docker image

Issue #754, opened 2022-04-27
Filter only specific status types in CLI


bkrein-vertex:
It'd be nice to be able to filter the output of the `polaris audit` CLI command to only display specific test results.  In other words I'd like to be able to only show tests that fail with a "Danger" status (ignoring "Warning", etc.).  

Perhaps adding flags like this?

- `--only-show-danger`
- `--only-show-warning`

Or something more complicated like this?

- `--show "Danger, Warning"`

Or a reusable flag?

- `--show Danger --show Warning`

sudermanjr:
Right now we have the `--only-show-failed-tests` which will get you part of the way there

sudermanjr:
I marked this as a an enhancement. If someone is willing to contribute, we would be happy to review!

jslivka:
This should now be supported with the introduction of https://github.com/FairwindsOps/polaris/pull/969

PR #969, opened 2023-06-27
Add option to filter audit results by severity level

jslivka:
This PR fixes #

## Checklist
* [x] I have signed the CLA

## Description

To address https://github.com/FairwindsOps/polaris/issues/754, I've chosen to implement verbosity options for displaying audit results in the same manner as log levels. `danger` is the least verbose option and will only include results of that severity. `warning` is the medium verbosity option and will only display results of severity `warning` or `danger`. By default, all results will be displayed. This will still function with `--only-show-failed-tests` and behave as expected.

Add verbosity options to the `audit` subcommand in the polaris CLI

Low verbosity (`danger`)

```
./polaris audit --audit-path=manifests --severity=danger | wc -l


Want more? Automate Polaris for free with Fairwinds Insights!
🚀 https://fairwinds.com/insights-signup/polaris 🚀
     116
```

Medium verbosity (`warning`, also implies inclusion of `danger`)

```
/polaris audit --audit-path=manifests --severity=warning | wc -l


Want more? Automate Polaris for free with Fairwinds Insights!
🚀 https://fairwinds.com/insights-signup/polaris 🚀
     268
```

### DIFF
--- a/pkg/validator/output.go
+++ b/pkg/validator/output.go
@@ -53,6 +53,26 @@ type AuditData struct {
 	Score                uint
 }
 
+// FilterResultsBySeverityLevel includes results according to the provided severity level:
+// 'danger' is the least verbose, 'warning' is medium verbosity, default behavior will
+// include all results, which currently also includes 'ignore'
+func (res AuditData) FilterResultsBySeverityLevel(severityLevel config.Severity) AuditData {
+	resCopy := res
+	resCopy.Results = []Result{}
+
+	filteredResults := funk.Map(res.Results, func(auditDataResult Result) Result {
+		return auditDataResult.filterResultsBySeverityLevel(severityLevel)
+	}).([]Result)
+
+	for _, result := range filteredResults {
+		if result.isNotEmpty() {
+			resCopy.Results = append(resCopy.Results, result)
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/validator/output.go
```diff
@@ -53,6 +53,26 @@ type AuditData struct {
...
+
+	filteredResults := funk.Map(res.Results, func(auditDataResult Result) Result {
+		return auditDataResult.filterResultsBySeverityLevel(severityLevel)
+	}).([]Result)
+
+	for _, result := range filteredResults {
+		if result.isNotEmpty() {
+			resCopy.Results = append(resCopy.Results, result)
+		}
+	}
```
rbren, line 71: Man 😓 the structure of the polaris output makes this harder than it should be.

Probably not for this PR, but if we wanted to do other kinds of filtering we could probably create a generic `AuditData.filter` method that takes in a `func(Result) bool`, so we only have to define the weird "drill into each result" logic in one place.

rbren (approved):
:awesome:

pkg/validator/output.go
jslivka: Yes that would be great, have some sort of filter function that takes a predicate

Issue #979, opened 2023-08-08
Problem with numbers in resource check


madiTG:
### What happened?

Hello

When creating object in kubernetes you are allowed to set limits and requests for CPU as number f.e. 2. 

```
    resources:
      limits:
        cpu: 2
        memory: 1024Mi
      requests:
        cpu: 1
        memory: 8Mi
```

```
limits:
  type: object
  properties:
     cpu:
       resourceMinimum: 100m
       resourceMaximum: "4"
```

When i do polaris audit vs a limit in range, but set as number it fails

### What did you expect to happen?

I would expect to parse numbers and compare them in the same manner as strings

### How can we reproduce this?

As written above. Check againsta a number vaule in resourceMinimum or resourceMaximum

### Version

8.4.0

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rbren:
I'm unable to reproduce this so far.

Just to be clear, you're setting `resourceMinimum` and `resourceMaximum` as strings, but setting the actual requests/limits using numbers, right?

Are you running against resources in-cluster, or as local YAML files?

rbren:
Here's my repro attempt: https://github.com/FairwindsOps/polaris/compare/rb/resource-range?expand=1

madiTG:
> I'm unable to reproduce this so far.
> 
> Just to be clear, you're setting `resourceMinimum` and `resourceMaximum` as strings, but setting the actual requests/limits using numbers, right?
> 
> Are you running against resources in-cluster, or as local YAML files?

Hello @rbren 

Yes, I AM setting these as strings, I couldn't find other option in docs. Should it work only string vs string and number vs number ? 

I launch polaris locally, during jenkins pipeline on yaml files.

madiTG:
My yaml file:

```
metadata:
  name: app
  labels:
    app: app
apiVersion: build.openshift.io/v1
kind: BuildConfig
spec:
  output:
    to:
      name: app:20230808-154210-49-app-
      kind: ImageStreamTag
  completionDeadlineSeconds: 1800
  resources:
    limits:
      cpu: 2
      memory: 1024Mi
    requests:
      cpu: 1m
      memory: 8Mi
  source:
    git:
      uri: https://gitlab/app.git
      ref: app
    type: Git
    dockerfile: |-
      #############
      ### build ###
      #############

  postCommit: {}
  strategy:
    type: Docker
[...]

PR #991, opened 2023-08-31
Fix numerical resource ranges

rbren:
This PR fixes an issue when parsing numerical resource ranges

Fixes issue https://github.com/FairwindsOps/polaris/issues/979

This one was really hard to track down. It actually doesn't trip up our tests, because we parse and validate known Kinds properly. The issue seems to be for unknown resources like the BuildConfig referenced in #979 

I've validated that the case provided in the issue now works.

## Checklist
* [x] I have signed the CLA

Support numerical CPU and memory amounts

Add a case to handle numbers


sudermanjr (approved):
LGTM. Love the tests

mggude (approved):

PR #998, opened 2023-09-26
chore: fix privilegeEscalationAllowed check

### LINKED ISSUE #997: privilegeEscalationAllowed forces label to be set
### What happened?

I want the test to fail only when the following configuration is detected:
```yaml
securityContext:
  allowPrivilegeEscalation: true
```
Currently, the test does fail when this configuration is present. However, I don't want it to fail when nothing is specified. Sometimes, it's not straightforward to specify this property, especially when using Helm Charts.

It's important to note that when the allowPrivilegeEscalation field is not explicitly set in the configuration, the default value is false. This default behavior is outlined in the Kubernetes documentation [here](https://github.com/kubernetes/design-proposals-archive/blob/main/auth/no-new-privs.md#changes-of-securitycontext-objects).

Therefore, the test should not require the property to be set, but rather, it should only fail when the property is explicitly set to true.

I anticipate that when no securityContext configuration is provided, this test should pass. This expected outcome is due to the default value of `allowPrivilegeEscalation` being set to `false`.

Create a deployment without specifying any `securityContext` settings, and then inspect the Polaris Dashboard. You will observe that the `privilegeEscalationAllowed` check fails. However, it's essential to note that, from Kubernetes' perspective, `allowPrivilegeEscalation` is indeed disabled by default in this scenario.

### Version

5.7.3

### Search

### Code of Conduct

### Additional context

_No response_

sempex:
This PR fixes #997 

## Checklist
* [x] I have signed the CLA

Fixes the Problem of this Issue #997 
I removed the requirement that the container must explicitly have the following configuration:
``` yaml
securityContext:
  allowPrivilegeEscalation: false
```
However, the check still fails when the configuration is as follows:
``` yaml
securityContext:
  allowPrivilegeEscalation: true
```
What's important to understand now is that when you omit any `securityContext` configuration entirely, the check will pass. This is because Kubernetes defaults to a configuration where `allowPrivilegeEscalation` is set to `false`.
-

### DIFF
--- a/checks/privilegeEscalationAllowed.yaml
+++ b/checks/privilegeEscalationAllowed.yaml
@@ -26,22 +26,13 @@ schema:
     properties:
       securityContext:
         $ref: "#/definitions/goodSecurityContext"
+  - properties:
       containers:
         type: array
         items:
           properties:
             securityContext:
               $ref: "#/definitions/notBadSecurityContext"
-  - properties:
-      containers:
-        type: array
-        items:
-          required:
-          - securityContext
-          properties:
-            securityContext:
-              $ref: "#/definitions/goodSecurityContext"
-
 mutations:
   - op: add
     path: /securityContext/allowPrivilegeEscalation


sempex:
The check is correct, have to fix go tests

PR #1025, opened 2024-01-03
FWI-5232 Conditional expressions should be at very top of additionalSchemaStrings

### LINKED ISSUE #1003: Issues with rolebindingClusterAdminClusterRole and rolebindingClusterRolePodExecAttach
### What happened?

I'm seeing some issues with the rolebindingClusterAdminClusterRole and rolebindingClusterRolePodExecAttach checks. I have IMO a simple ServiceAccount with a simple Role and RoleBinding, but Polaris throws an error with the aforementioned checks and I don't fully understand how to fix them nor if they are an issue. It seems more like an bug of Polaris.

Trying to check deeper I found this link https://polaris.docs.fairwinds.com/checks/security/, where it states:
- rolebindingRolePodExecAttach - Fails when the RoleBinding references a ClusterRole that allows Pods/exec, allows pods/attach, or that does not exist.
- rolebindingClusterAdminClusterRole - Fails when the RoleBinding references the default cluster-admin ClusterRole or one with wildcard permissions.

And to my eyes, none of them apply to my conf, this is a simplified version but it triggers the issue:
```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: some-api
  namespace: dev
  labels:
    role: service
    tier: backend
imagePullSecrets:
  - name: regcred
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: some-api
  namespace: some-ns
rules:
  - apiGroups: [""]
    resources: [services]
    verbs: [get, list]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: some-api
  namespace: some-ns
roleRef:
  kind: Role
  name: some-api
  apiGroup: rbac.authorization.k8s.io
subjects:
  - kind: ServiceAccount
    name: some-api
    namespace: dev
```

To show success for my configuration.

Put the configuration I shared in a test.yaml file and run: `polaris audit --audit-path test.yaml --format=pretty`

### Version

8.5.1

### Search

### Code of Conduct

### Additional context

_No response_

jslivka:
This PR fixes #1003

## Checklist
* [x] I have signed the CLA

## Description

Polaris checks utilize `additionalSchemaStrings`, an additional section of jsonschema that will only evaluate if the object conforms to `schemaString`. The conditional statements in `additionalSchemaStrings` need to be arranged in a manner such that when there's nothing to evaluate, the string is empty (not the object therein).

Fix false positives reported when running the Polaris CLI for the following checks:

* clusterrolebindingClusterAdmin
* rolebindingClusterAdminClusterRole
* rolebindingClusterRolePodExecAttach
* rolebindingRolePodExecAttach

adjust the ordering of conditional expressions in `additionalSchemaStrings`:

e.g.

```
 additionalSchemaStrings:
   rbac.authorization.k8s.io/ClusterRole: |
-    type: object
-    # This schema is validated for all roleBindings, regardless of their roleRef.
     {{ if eq .roleRef.kind "ClusterRole" }}
     {{ if and (not (hasPrefix .metadata.name "system:")) (ne .metadata.name "gce:podsecuritypolicy:calico-sa") }}
+    # This schema is validated for all roleBindings, regardless of their roleRef.
+    type: object
```

so that there is not an empty object being evaluated, but rather an empty string

I've also added success test cases as a regression check

### DIFF
--- a/checks/clusterrolebindingClusterAdmin.yaml
+++ b/checks/clusterrolebindingClusterAdmin.yaml
@@ -37,10 +37,10 @@ schemaString: |
                 const: "cluster-admin"
 additionalSchemaStrings:
   rbac.authorization.k8s.io/ClusterRole: |
-    type: object
-    # Do not alert on default ClusterRoleBindings.
     {{ if (ne .roleRef.name "view") }}
     {{ if and (ne .metadata.name "cluster-admin") (not (hasPrefix .metadata.name "system:")) (ne .metadata.name "gce:podsecuritypolicy:calico-sa") }}
+    # Do not alert on default ClusterRoleBindings.
+    type: object
     required: ["metadata", "rules"]
     allOf:
       - properties:
--- a/checks/rolebindingClusterAdminClusterRole.yaml
+++ b/checks/rolebindingClusterAdminClusterRole.yaml
@@ -45,10 +45,10 @@ schemaString: |
                 const: "cluster-admin"
 additionalSchemaStrings:
   rbac.authorization.k8s.io/ClusterRole: |
-    type: object
-    # This schema is validated for all roleBindings, regardless of their roleRef.
     {{ if eq .roleRef.kind "ClusterRole" }}
     {{ if and (not (hasPrefix .metadata.name "system:")) (ne .metadata.name "gce:podsecuritypolicy:calico-sa") }}
+    # This schema is validated for all roleBindings, regardless of their roleRef.
+    type: object
     required: ["metadata", "rules"]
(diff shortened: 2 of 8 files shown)


rbren:
For posterity, here's the line that makes this work: https://github.com/FairwindsOps/polaris/blob/0b765df04db2960bb744ab3dcb23fdbe304cbcde/pkg/config/schema.go#L243

rbren (approved):

PR #1053, opened 2024-06-26
INSIGHTS-90 - implement HPA minAvailable and HPA maxAvailable checks

vitorvezani:
This PR fixes internal INSIGHTS-90

## Checklist
* [X] I have signed the CLA

- implement support for HPA `minAvailable` check
- implement support for HPA `maxAvailable` check

- I had to implement a custom validator function for validations that are not supported by JSON validation
  - i.e.: in HPA `maxAvailable` check, we need to check fix `maxAvailable` is greater than `minAvailable` and in JSON validation you cannot self-reference fields

### DIFF
--- a/pkg/config/checks/hpaMaxAvailability.yaml
+++ b/pkg/config/checks/hpaMaxAvailability.yaml
@@ -1,4 +1,4 @@
 successMessage: HPA has a valid max and min replica configuration
-failureMessage: HPA maxReplicas should be greater than minReplicas
+failureMessage: HPA maxReplicas and minReplicas should be different 
 category: Reliability
 target: autoscaling/HorizontalPodAutoscaler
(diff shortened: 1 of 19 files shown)


pkg/config/checks/hpaMaxAvailability.yaml
```diff
@@ -0,0 +1,4 @@
+successMessage: HPA has a valid max and min replica configuration
+failureMessage: HPA maxReplicas should be greater than minReplicas
+category: Reliability
+target: autoscaling/HorizontalPodAutoscaler
```
vitorvezani, line 4: Since no schema is provided, it will use the function validator defined by the same checkID (`hpaMaxAvailability`)

sudermanjr (approved):

PR #1070, opened 2024-10-02
Fix: rolebindingRolePodExecAttach check

greg5813:
Fix the case of a RoleBinding that points to a ClusterRole.
In that case, we ignore the RoleBinding validation altogether since it will be evaluated by the rolebindingClusterRolePodExecAttach check.

This PR fixes the rolebindingRolePodExecAttach check

## Checklist
* [X] I have signed the CLA

Fix the check of a RoleBinding that references a Role allowing pods/exec or pods/attach in the case that it points to a ClusterRole. 

Modify the kind field value of the roleRef field of the RoleBinding from "Role" to "ClusterRole" in the schemaString

NA

### DIFF
--- a/pkg/config/checks/rolebindingRolePodExecAttach.yaml
+++ b/pkg/config/checks/rolebindingRolePodExecAttach.yaml
@@ -17,7 +17,7 @@ schemaString: |
               const: "rbac.authorization.k8s.io"
             kind:
               type: string
-              const: "Role"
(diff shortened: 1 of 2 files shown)


vitorvezani:
Thanks for your contribution Greg, I've added a test to avoid regressions in the future

sudermanjr (approved):
LGTM. Thanks!!

PR #1200, opened 2026-05-21
Fixes the Polaris download URL

### LINKED ISSUE #1199: bug: get_polaris.sh script fails to download release due to asset naming mismatch
### Description
The `get_polaris.sh` script used in GitHub Actions fails to download the Polaris binary because the release asset naming convention changed, causing a `404 Not Found` error during download.

### Current Behavior
The script currently constructs the URL using the following pattern:
`POLARIS_URL=https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_linux_amd64.tar.gz`

However, recent releases use the version tag (without the "v" prefix) inside the archive filename itself (e.g., `polaris_10.2.0_linux_amd64.tar.gz`). 

Because of this mismatch, the binary is never downloaded or extracted, leading to the following error in GitHub Workflow logs:

```bash
/get_polaris.sh: line 7: polaris: command not found
Installing polaris to path from  https://github.com/FairwindsOps/polaris/releases/download/v10.2.0/polaris_linux_amd64.tar.gz
  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed

  0      0   0      0   0      0      0      0                              0
100      9 100      9   0      0     57      0                              0
100      9 100      9   0      0     57      0                              0
100      9 100      9   0      0     57      0                              0
gzip: invalid magic
tar: Child returned status 1
tar: Error is not recoverable: exiting now

```
Image

### Expected Behavior
The script should dynamically strip the "v" prefix from the `$INPUT_VERSION` variable when constructing the asset filename, ensuring it fetches the correct archive.

### Suggested Fix
We can use `tr` to strip the "v" character from the version string for the filename part of the URL:

```bash
POLARIS_URL=[https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_$(echo](https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_$(echo) -n $INPUT_VERSION | tr -d "v")_linux_amd64.tar.gz

### Steps to reproduce:
1. Use the Polaris GitHub Action in a workflow with any recent version tag (e.g., `v10.2.0` or custom version input).
2. Run the workflow.
3. The step will fail during execution because the script cannot find the `polaris` binary.

### Example Workflow step that triggers the bug:
```yaml
      - name: Setup polaris
        with:
          version: v10.2.0

### Version

10.2.0

### Search

- [x] I did search for other open and closed issues before opening this.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by piar1989**
This PR fixes #1199

## Checklist
* [x] I have signed the CLA

The goal of this PR is to fix a bug in `get_polaris.sh` where the script fails to download the Polaris binary for recent releases due to a mismatch in the release asset naming convention.

I updated the `POLARIS_URL` construction inside `get_polaris.sh`. It now dynamically strips the "v" prefix from the `$INPUT_VERSION` variable using `tr -d "v"` specifically for the asset filename part of the URL, while keeping the full tag for the download path.

None. This is the most straightforward and lightweight way to handle the version string adjustment directly inside the Bash script without adding external dependencies.

### DIFF
--- a/.github/actions/setup-polaris/get_polaris.sh
+++ b/.github/actions/setup-polaris/get_polaris.sh
@@ -3,7 +3,7 @@ if [[ -z "$INPUT_VERSION" ]]; then
   echo "Missing polaris version information"
   exit 1
 fi
-POLARIS_URL=https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_linux_amd64.tar.gz
+POLARIS_URL=https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_$(echo -n $INPUT_VERSION | tr -d "v")_linux_amd64.tar.gz
 polaris version | grep "$INPUT_VERSION" &> /dev/null
 if [ $? == 0 ]; then
    echo "Polaris $INPUT_VERSION is already installed! Exiting gracefully."

### CODE REVIEW

**Review by vitorvezani — CHANGES_REQUESTED** [review]
Thanks for the contribution @piar1989, to be backwards compatible, I think we will need something like this
```
if [[ "$INPUT_VERSION" == v* ]]; then
  POLARIS_URL=https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_$(echo -n $INPUT_VERSION | tr -d "v")_linux_amd64.tar.gz
else
  POLARIS_URL=https://github.com/FairwindsOps/polaris/releases/download/$INPUT_VERSION/polaris_linux_amd64.tar.gz
fi
```

So that if old patterns are used, we still resolve to correct URL, i.e.:
- `https://github.com/FairwindsOps/polaris/releases/download/4.2.0/polaris_linux_amd64.tar.gz`

PR #1230, opened 2026-08-24
ci: remove deprecated fairwinds/oss-docs orb to unblock Renovate

Ranakghosh7:
The Renovate bot's dependency dashboard is failing because the fairwinds/oss-docs@0 orb is returning a no-result error from the CircleCI registry.

I have removed the deprecated orb declaration from .circleci/config.yml and deleted the unused oss-docs/publish-docs job from the release workflow. Validated locally using circleci config validate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   rok8s: fairwinds/rok8s-scripts@16.0.0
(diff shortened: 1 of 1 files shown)


sudermanjr:
This will be fixed in a larger refactoring of how we manage ci config centrally and internally. Thanks for bringing it up though!

